Issue #2385, opened 2026-05-05
epic: adopt addyosmani/agent-skills patterns + skills (consensus 100%)


williamzujkowski:
## Source

Research-then-adopt of [addyosmani/agent-skills](https://github.com/addyosmani/agent-skills) (27.8k stars, MIT-licensed, "Production-grade engineering skills for AI coding agents").

Consensus vote (2026-05-04, simple_majority quickMode): **APPROVED 100%** — Software Architect 0.82, Security Engineer 0.85, Scope Steward timeout. Tier-level refinements from architect incorporated below.

## Distinctive ideas worth importing

1. **Anti-rationalization tables** — pre-emptive `excuse → counter` rebuttals.
2. **Verification-first checklists** at end of every skill.
3. **Red-flags sections** — symptoms that signal the skill is being misapplied.
4. **References loaded on-demand** (accessibility, performance, security, testing, orchestration).
5. **Phase taxonomy**: DEFINE → PLAN → BUILD → VERIFY → REVIEW → SHIP for skill organization.
6. **Google engineering concepts**: Hyrum's Law, Chesterton's Fence, Beyoncé Rule, DAMP-over-DRY.

### Tier A — Adopt 7 NEW skills we lack (was 10; 3 dropped per architect)

- `test-driven-development`
- `code-simplification`
- `deprecation-and-migration` (would have helped epic #2368)
- `performance-optimization`
- `api-and-interface-design`
- `browser-testing-with-devtools` *(security note: URL allowlist + untrusted-DOM handling per Security voter)*
- `context-engineering`

**Dropped from Tier A** (architect: overlap with existing skills; fold their distinctive content into Tier D cross-pollination instead):

- ~~`spec-driven-development`~~ → cross-pollinate into `dev-pipeline` (we have `execute_spec` MCP tool already)
- ~~`idea-refine`~~ → cross-pollinate into `requirements-gathering`
- ~~`incremental-implementation`~~ → cross-pollinate into `implement-feature`

### Tier B — Adopt 5 references in `skills/references/` (load-on-demand)

- `accessibility-checklist.md`
- `performance-checklist.md`
- `security-checklist.md`
- `testing-patterns.md`
- `orchestration-patterns.md`

### Tier C — Adopt 3 agent personas as **subagent prompt templates** (architect: NOT as new voter roles to avoid 7-role panel inflation)

- `code-reviewer.md` → `.claude/agents/code-reviewer.md`
[...]

PR #2389, opened 2026-05-05
feat(skills): tier A2 + trigger whitespace fix (#2385)

williamzujkowski:
## Summary

Tier A2 of epic #2385 (consensus-approved 100% adoption of addyosmani/agent-skills). Adds 4 more skills and patches \`scripts/generate-skills-index.ts\` to fix the YAML-newline trigger-extraction bug that bit PR #2386.

### \`performance-optimization\`

Measure-first MIFVG cycle (Measure → Identify → Fix → Verify → Guard). Anti-rationalization for premature optimization excuses ("Big-O says it's faster," "JavaScript engines optimize this," "We'll measure in production").

### \`api-and-interface-design\`

Hyrum's Law, contract-first, validate-at-boundaries (cross-references our zero-\`any\` policy + \`.rules/untrusted-input.md\`), consistent error semantics (Result-typed canonical, never mix patterns), discriminated unions, branded IDs, input/output separation. Cross-links to \`deprecation-and-migration\` for the safe-evolve path.

### \`browser-testing-with-devtools\`

**Per Security voter's epic-vote concern**: explicit URL-allowlist + untrusted-DOM handling. Browser content (DOM, console, network, JS-execution output) treated as untrusted by default. No instruction-following from page content. JS-execution constrained to read-only state inspection — no \`fetch\`, no \`document.cookie\`, no exfiltration.

### \`context-engineering\`

Six-level context hierarchy (rules → memory → spec → source → live state → conversation). Subagent fan-out discipline (3-4 wave, < 500-word prompts, output budget, status line). Confusion-management pattern (surface ambiguity, don't silently choose). Inline-planning pattern.

## Script fix

\`scripts/generate-skills-index.ts\`: trigger phrases extracted from YAML literal-block descriptions could contain literal newlines if the description wrapped at column 80. Those newlines then propagated into CLAUDE.md's skill table cells, breaking the markdown table (MD038 + MD056). Now normalized via a new \`normalizeTrigger()\` helper applied to all three trigger-extraction paths.

This was the root cause of the markdown-lint failure on PR #2386, which had to be fix-forwarded by PR #2388. Now fixed at the source — future skills won't hit this regardless of how their YAML wraps.

## Verification

- \`pnpm typecheck\` clean
- \`pnpm lint\` clean
- \`pnpm test\`: 26,386 passed, 16 skipped (1054 files)
- \`npx markdownlint\` clean (this would have failed without the script fix)
- \`npx tsx scripts/check-docs-indexed.ts\`: 118 indexed, 0 unindexed

## Test plan

- [x] Local typecheck/lint/test/markdownlint all green
- [x] Trigger normalization verified by re-running governance after intentionally-wrapped descriptions
- [ ] CI green
- [ ] \`consensus_vote\` QA approves

Refs #2385


williamzujkowski:
Addressed architect's non-blocking suggestion: added a regression test in `scripts/generate-skills-index.test.ts` that asserts no trigger contains internal newlines, double-spaces, or tabs. This is the contract `normalizeTrigger()` establishes; the test catches any future YAML-literal-block-wrap bug at the index-generation layer instead of relying on downstream markdownlint of the generated CLAUDE.md table.

Verified: 7 tests pass.

Issue #2398, opened 2026-05-05
epic: adopt nexu-io/open-design process patterns (self-critique, brand extraction, decision batching)


williamzujkowski:
## Source

Research-then-adopt of [nexu-io/open-design](https://github.com/nexu-io/open-design) (24.3k stars, **Apache-2.0**, "open-source alternative to Claude Design").

## Scope decision

Open Design is a **complete design platform** (Next.js web app + Electron desktop + Express daemon + 19 design-specific skills + 71 design systems + HyperFrames video framework + sandboxed preview + PDF/PPTX export). 95% of that surface is **out of scope** for nexus-agents — we're a multi-agent governance substrate, not a design tool.

The **process patterns**, however, are domain-general and high-leverage. This epic adopts those, NOT the application infrastructure.

**Explicitly skip:**
- All 19 design-specific skills (web-prototype, saas-landing, mobile-app, hatch-pet, hyperframes, etc.) — out of charter
- 71 design systems (DESIGN.md catalog) — too much content; reference upstream if needed
- Daemon, Electron, Next.js app, sidecar protocol — entirely orthogonal architecture
- HTML/PDF/PPTX/MP4 export — different product
- HyperFrames video — separate library

### Tier 1 — Five-dimensional self-critique (highest leverage)

Adopt the generalizable pre-emit critique pattern from `skills/critique/SKILL.md`:

- 5 dimensions, each scored 0–10 with bands (0–4 Broken / 5–6 Functional / 7–8 Strong / 9–10 Exceptional)
- **Evidence-citation discipline** — "scored 4 because [specific element]" beats "feels inconsistent"
- **Don't average up** rule — score is the *worst sustained band*
- **Don't grade-inflate** — 7 means *strong*, not *acceptable*
- Pre-emit self-check: agent runs critique on its own output before emitting; anything <3/5 is a regression — fix and rescore

**Adoption shape**: New skill `self-critique` (generic, NOT design-specific). The 5 dimensions adapt per task type:

- **Code review**: Correctness / Readability / Architecture / Security / Performance (already in `reviewing-code` skill — wire critique pattern to it)
- **Design**: Philosophy / Hierarchy / Detail / Functionality / Innovation (Open Design's original)
- **Docs**: Accuracy / Discoverability / Density / Examples / Tone
- **Spec/PR**: Completeness / Testability / Reversibility / Stakeholder-fit / Scope
[...]

PR #2400, opened 2026-05-05
feat(skills): tier 2+3 — brand extraction + DESIGN.md schema (#2398)

williamzujkowski:
## Summary

Bundled Tiers 2 + 3 of epic #2398. Both target the same skill (`ui-ux-design`); shipping together avoids fragmenting the cross-references.

## Tier 2 — Brand extraction protocol

5-step protocol adopted from Open Design's brand-extraction discipline. Per security voter's QA on the epic vote: agent-driven URL fetch is SSRF / DoS / path-traversal risk. The protocol is **gated on explicit safety guards**:

1. **Locate** — local repo asset preferred, user-pasted excerpt as fallback, external URL as last resort
2. **Safety guards (when URL is the only path)** — non-negotiable:
   - Explicit user confirmation (never auto-fetch)
   - HTTPS only (reject `http://`, `file://`, `ftp://`, protocol-relative `//host`)
   - **Public-IP allowlist**: reject RFC 1918 + link-local + CGNAT + multicast + IPv6 equivalents (enumerated inline with CIDRs)
   - Content-type allowlist (HTML / CSS / SVG / PNG / JPEG / WebP only)
   - 5 MB size cap, 30 s timeout
   - Treat fetched content as untrusted per `.rules/untrusted-input.md`
3. **Extract tokens** — concrete `grep -hoiE` patterns
4. **Codify in `brand-spec.md`** — path-traversal guard (cwd subtree only)
5. **Vocalize** — read tokens back to user before generating code

## Tier 3 — 9-section DESIGN.md schema

Portable design-system schema (cross-tool: Open Design, Claude Design, future nexus-agents UI tooling). Sections: Visual theme / Color palette / Typography / Component stylings / Layout / Depth & elevation / Dos and don'ts / Responsive strategy / Agent prompt guide.

## Bundled — 8-dimension brief input format

Adopted from upstream's `design-brief` skill. Structured brief schema (palette / accent / typography / display / layout / mood / density / exclude) with default-resolution rules. **"Don't silently default"** discipline — surface unspecified dimensions for confirmation.

## Tier 4 skipped

Audit of severity language across skills found no drift — `critical/high/medium/low` (CVSS-aligned for security) and `P1/P2` (issue-priority for backlog) coexist appropriately. Forcing one canonical scheme would be over-engineering.

## License

Apache-2.0 attribution in section quotes. Compatible with our MIT.

## Verification

- `npx markdownlint 'skills/**/*.md'` clean
- `npx vitest run scripts/generate-skills-index.test.ts`: 7/7 pass
- Skill count unchanged at 26 (additive to existing skill, not a new one)

## Test plan

- [x] Local lint + test green
- [ ] CI green
- [ ] `consensus_vote` QA approves (especially the URL-fetch guards)

Refs #2398

### DIFF
--- a/.changeset/2398-tier2-3-brand-extraction.md
+++ b/.changeset/2398-tier2-3-brand-extraction.md
@@ -1,0 +1,27 @@
+---
+'nexus-agents': patch
+---
+
+Tiers 2 + 3 of epic #2398 — enhance `ui-ux-design` skill with patterns from Apache-2.0-licensed [nexu-io/open-design](https://github.com/nexu-io/open-design):
+
+**Tier 2 — Brand extraction protocol** (5 steps with explicit safety guards per security voter):
+
+1. **Locate** — local repo asset preferred, user-pasted excerpt as fallback, external URL as last resort
+2. **Safety guards (when fetching URL)** — non-negotiable per security review:
+   - Explicit user confirmation (never auto-fetch)
+   - HTTPS only (reject `http://`, `file://`, `ftp://`, protocol-relative)
+   - Public-IP allowlist (reject RFC 1918 + link-local + CGNAT + IPv6 equivalents — full list inline)
+   - Content-type allowlist (HTML/CSS/SVG/PNG/JPEG/WebP only)
+   - 5 MB size cap, 30 s timeout
+   - Treat fetched content as untrusted per `.rules/untrusted-input.md`
+3. **Extract tokens** — concrete `grep -hoiE` patterns for hex codes, font families, spacing scale
+4. **Codify in `brand-spec.md`** — path-traversal guard (cwd subtree only)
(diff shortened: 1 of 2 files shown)


williamzujkowski:
Addressed two non-blocking notes from QA:

1. **Security**: ReDoS concern on grep over 5 MB untrusted input. Patterns shown are simple character classes without nested quantifiers, but added an explicit note recommending ripgrep (RE2, no backtracking) and listing dangerous pattern shapes to avoid (`(a+)+`, `(a|a)+`, `^(a+)+$` over user-controlled text).
2. **Architect**: DNS-rebinding gap if this becomes a live fetch tool. Added requirement for **resolve-then-pin-IP** semantics — resolve hostname once, validate IP, connect to that IP directly. Otherwise an attacker can return a public IP at validation time and 127.0.0.1 on the actual connection.

PR #3309, opened 2026-06-01
refactor(config): adopt parseBoolEnv in 4 duplicate bool-env parsers (#3297)

williamzujkowski:
## What

The clean, verifiable slice of #3297 (env-parser collapse). Four sites reimplemented `parseBoolEnv`'s exact logic (`process.env[K] === '1' || === 'true'`) inline; they now adopt the **existing canonical** helper:

- `cli/research-scaffold.ts` `scaffoldDisabled`
- `cli/hooks/handlers/handler-utils.ts` `isFeatureDisabled` + `isVerboseLogging`
- `adapters/sdk/custom-api-validation.ts` `resolveAllowPrivateFromEnv`

`parseBoolEnv` previously had **no importers** — this gives the canonical helper its first real consumers.

## Behavior

Preserving except a **desirable normalization**: these flags become case-insensitive (`TRUE`/`True` now parse as true, matching every other `parseBoolEnv` flag). The **68 existing tests** for these sites pass unchanged.

## Scope (per #3297 verify-first)

Only the 4 sites whose semantics already matched (`'1'|'true'`) are adopted. Left untouched (divergent semantics — would be behavior changes, not consolidation):
- `'true'`-only key reads (cli-server-auth, cli-server-tools, setup-formatting, v2-config, reflective-retriever) — adopting would *widen* to also accept `'1'`.
- `step-notifications` accepts `'on'` (parseBoolEnv doesn't).
- `portable-mode` needs its explicit tri-state (explicit-false vs unset).

These remaining normalizations are a separate, deliberate semantics decision — noted on #3297.

Part of #3297 / epic #3288.

### DIFF
--- a/.changeset/adopt-parseboolenv.md
+++ b/.changeset/adopt-parseboolenv.md
@@ -2,11 +2,11 @@
 'nexus-agents': patch
 ---
 
-refactor(config): adopt canonical parseBoolEnv in 4 duplicate bool-env parsers (#3297)
+refactor(config): adopt canonical parseBoolEnv in 3 duplicate bool-env parsers (#3297)
 
-Four sites reimplemented `process.env[K] === '1' || === 'true'` inline (research
-scaffold, two hook-utils flags, custom-API allow-private). They now call the
-existing canonical `parseBoolEnv(key, false)`. Behavior-preserving except env
-flags become case-insensitive (a desirable normalization — `TRUE` now parses as
-true). The `'true'`-only and `'on'`-accepting sites are left as-is (different
-semantics); portable-mode keeps its explicit tri-state.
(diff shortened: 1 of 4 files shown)


williamzujkowski:
Revised: **excluded the SSRF-guard-bypass flag** (`NEXUS_CUSTOM_API_ALLOW_PRIVATE` in custom-api-validation) from the parseBoolEnv adoption. That flag disables an SSRF check, so it stays case-sensitive/fail-closed — widening it to case-insensitive via parseBoolEnv would let extra case variants (e.g. `TRUE`) loosen a security control. Only the 3 benign flags (research-scaffold, handler-utils ×2) adopt the canonical helper. 68 tests pass.

Issue #4449, opened 2026-08-15
tech-debt(docs): TypeDoc markdown regeneration churns ~40k lines because permalinks embed the commit SHA — 5 relevant lines out of 39,855


williamzujkowski:
## Summary

The committed TypeDoc markdown under `docs/api/` cannot be kept in sync, because a regeneration rewrites ~40,000 lines regardless of what changed. The dominant cause is that every `Defined in:` permalink embeds the **current commit SHA**, so every link in every file changes on every commit.

## Measurement (2026-08-15, from a PR that removed exactly three symbols)

Ran `pnpm --filter nexus-agents docs:api:md` on a branch whose only source change was deleting `WorkBalancer`, `getCapacityDashboard`, and `fetchCapacityData`:

| Metric | Count |
| --- | --- |
| Total changed lines in `docs/api/` | **39,855** |
| Lines that are `Defined in:` commit-SHA permalinks | 14,438 |
| Lines added with trailing whitespace | 450 |
| `title:` frontmatter quote-style flips (`'` → `"`) | 40 |
| Files touched | 20 |
| **Lines actually mentioning a removed symbol** | **5** |

Signal-to-noise: **5 relevant lines out of 39,855**, or 0.013%.

Example of the dominant churn — same file, same line number, only the SHA differs:

```diff
-Defined in: [.../step-executor.ts:101](https://github.com/.../blob/5c12a37f0775ec3c299142f22bd39afd955bbbaf/...#L101)
+Defined in: [.../step-executor.ts:101](https://github.com/.../blob/1f90f2afdb7e830988823e4257520c495dfc282b/...#L101)
```

## Relationship to #2027

#2027 observed this pain and resolved it by making the `TypeDoc Verification` gate **non-blocking** (warning only). That was the right call for unblocking PRs, but it treated the symptom: the gate stopped complaining, and the committed docs have been drifting ever since with nothing to correct them.

#2027's hypothesis was new public exports and Zod enum ordering. The measurement above shows the real driver is the commit SHA in permalinks, which guarantees ~100% churn on **every** commit, whether or not any export changed. That is a different problem with a different fix.

## Consequence

Committing these files is currently a no-op ritual: nobody can review a 40k-line diff, so nobody regenerates, so the docs stay stale. `docs/api/cli-adapters.md` documents `getCapacityDashboard()` right now, after that method has been deleted, and I deliberately left it stale in the deletion PR rather than attach 40k lines of unrelated churn to it.

[...]

Issue #4373, opened 2026-08-09
feat(cli-adapters): normalized capacity_exhausted diagnostic + capacity-aware routing exclusion (#4351 criterion 3) — exclusion held on #4456


williamzujkowski:
Child of #4351 (criterion 3). Split out after auditing the acceptance criteria against `e4d1cc929f` — the terminal-state increments (#4362, #4363) do not touch this.

## What is missing

**No normalized capacity diagnostic exists.** `grep -rn "capacity_exhausted"` returns nothing tree-wide. `CliErrorCode` (`cli-adapters/types-core.ts:108-118`) carries `RATE_LIMITED` / `BUDGET_EXCEEDED` / `UNKNOWN`, none of which distinguishes "this key is out of weekly quota" from "slow down, you are being throttled". `RATE_LIMIT_PATTERNS` (`orchestration/outcomes/outcome-types.ts:146-154`) has no weekly-quota or context-window pattern, so the outcome taxonomy cannot classify it either.

**Capacity is tracked but never routed on.** `CompositeRouter.getCapacityDashboard()` (`cli-adapters/composite-router.ts:1028`) is read-only. Nothing in `composite-router.ts` or `composite-router-helpers.ts` consults `CapacityStatus.exhausted` or `getCapacity()` to exclude an arm or trigger failover. The capacity tracker exists, has tests, and has no routing consumer.

## What already landed, so scope this correctly

#4330 (PR #4370) fixed the **CLI voter-panel** half by a different mechanism: `BaseCliAdapter` now feeds `defaultCliCircuitBreakerRegistry`, so a rate-limited CLI trips its breaker and `isCliServingForVoters` drops it from the next panel. That is failover-by-circuit-breaker, driven by observed failures.

This issue is the complementary path: excluding an arm *before* it fails, from capacity the tracker already holds, and giving the failure a diagnostic specific enough for a caller to act on. Do not duplicate the breaker mechanism — the two should compose, not race.

## Acceptance

- A normalized capacity-exhaustion code/category, distinct from generic rate-limiting, produced by the adapters that can detect it
- `CompositeRouter` excludes an exhausted arm, with a test proving exclusion (not just that the dashboard reports it)
- No-quorum behaviour when too few healthy voters remain after exclusion
- An adapter-level test for the authenticated-but-weekly-exhausted case — the state that produced the #4351 report, where every voter returned `stop_sequence` with zero tokens and the panel still reported healthy capacity

## Related

[...]

Issue #2027, opened 2026-04-19
tech-debt: TypeDoc drift failures on PRs touching execute-expert/orchestrate


williamzujkowski:
## Observation

Three PRs in the past session (#2021, #2024, #2026) all hit the \`TypeDoc Verification\` CI gate with trivial drift requiring a manual regen commit. Each time the drift was limited to small HTML changes in generated docs for functions I touched.

## Hypothesis

TypeDoc's HTML output is sensitive to:

1. Any new public export (the \`deriveXxxPolicy\` helpers I added were \`async function\`, not exported — but the regen still touched nearby \`.html\` files in the same module)
2. Inferred Zod enum ordering (SkillCapabilitiesSchema.html has drifted on unrelated PRs before per git log)

## Why it's painful

Every PR touching an orchestrate/execute-expert helper needs an extra round-trip:
1. Push PR
2. Wait for TypeDoc Verification to fail
3. Run \`pnpm --filter nexus-agents docs\` locally
4. Commit the regenerated HTML
5. Re-trigger CI

This adds ~5-10 minutes to every merge.

## Options

1. **Drop the gate** — TypeDoc drift is a low-severity issue. Make the CI job warning-only instead of blocking.
2. **Fix determinism** — investigate whether newer TypeDoc versions or different options produce stable output for our enum/schema shapes.
3. **Auto-regenerate in CI** — on drift detection, have CI commit the regen back (as \`github-actions[bot]\`) rather than failing the PR.

Lowest effort: option 1. Least risky: option 3 with a bot PR instead of direct commit. Option 2 is the right long-term fix.

## Acceptance

- Either TypeDoc drift no longer blocks merges OR the regen happens automatically.
- Contributors don't have to learn about TypeDoc regeneration as a side-ritual of landing unrelated PRs.

PR #4450, opened 2026-08-15
chore(routing): delete WorkBalancer and getCapacityDashboard (7/0 vote)

### LINKED ISSUE #4378: chore(routing): delete WorkBalancer + getCapacityDashboard (7/0 vote) — unblocks #4373
Found while auditing capacity consumers for #4374/#4373. Filed separately because deciding their fate is not in scope for either.

## What is unwired

**`context/work-balancer.ts`** implements real capacity-based decisions:
- `getCapacityScore` (`:267-272`) zeroes an arm when `capacity.exhausted`
- `getNextQueuedTask` (`:312-318`) admits a task only if `estimatedTokens <= remainingTokens`
- `calculateCapacityRatio` (`:378-379`)

`WorkBalancer` is **never instantiated in production**. The only non-test references in the tree are its own module, `work-balancer-types.ts`, and the `context/index.ts` re-export.

**`CompositeRouter.getCapacityDashboard()`** (`cli-adapters/composite-router.ts:1028`) has **zero production callers** — referenced only by its own interface declaration (`:180`) and two test mocks (`mcp/tools/delegate-to-model.test.ts:425`, `learning/feedback-integration.test.ts:75`).

The only live consumer of `CapacityStatus` anywhere is `doctor`, for display.

## History — speculative, not interrupted

Per the research-intent rule, under-wired code can be interrupted work. Checked: it is not.

`work-balancer.ts` was added 2026-01-05 in `89085ad31c` "feat: implement Phase 2-4 infrastructure components" — a bulk commit adding **37 files** at once. Every subsequent commit touching it is a cross-cutting sweep unrelated to capacity (`0f78327b07` determinism migration, `ec635cf907` clamp consolidation). In seven months no commit has attempted to wire it. `getCapacityDashboard` has the same profile — its only history entry is a registry-centralization refactor.

## The decision, and why it is not obvious

This is not a straightforward delete. `WorkBalancer` carries a task-queueing model that a future scheduler could legitimately want, and the repo's capability-bias principle explicitly permits building ahead of demand **when a named consumer or near-term loop will measurably use it**. The question is whether one exists now.

Options:
1. **Wire it** — if capacity-aware routing (#4373) should consume it rather than growing a parallel implementation. This is the outcome that would make it worth having kept.
2. **Delete both** — no named consumer after seven months; re-derive from git if a scheduler ever needs it.
3. **Keep, documented as unwired** — the worst option unless a specific near-term loop is named, since it reads as live infrastructure to anyone auditing capacity handling. It already cost one audit cycle.

The forcing function is #4373: if capacity-aware exclusion gets built without settling this, the result is a second capacity-decision implementation alongside a dormant one — precisely the anti-sprawl violation the canonical-paths rule prohibits.

**Blocked by:** nothing. **Blocks:** the routing half of #4373. Decide before that lands.

williamzujkowski:
Closes #4378. PR 1 of 2 in the sequence decided by the vote; #4373 follows.

## Decision

Live 7-voter `higher_order` `consensus_vote`, **7/0 unanimous**, no simulated votes, record persisted. Full reasoning on #4378.

**This overturned my own initial recommendation.** I had leaned toward *wiring* `WorkBalancer` into #4373 rather than deleting it, on the grounds that it already encoded the capacity semantics and the canonical-paths rule discourages parallel implementations. The panel inverted that reading, and verification confirmed they were right.

## Why delete rather than wire

`WorkBalancer` is not a queue that happens to know about capacity — it is a second routing implementation. `work-balancer.ts:250-264` carried its own weighted capability scoring:

```ts
const weights = this.calculateCapabilityWeights(task);
const weightedScore =
  reasoningMatch * weights.reasoning +
  codeGenMatch   * weights.codeGeneration +
  speedMatch     * weights.speed +
  costMatch      * weights.cost +
  contextScore   * weights.context;
```

That duplicates what `SharedTaskAnalyzer` + `TopsisRouter` own canonically. Injecting it into `CompositeRouter` would have **committed** the anti-sprawl violation it was invoked to prevent — 388 lines of queue-and-scoring machinery kept alive to reuse ~6 lines of predicate.

The research-intent rule required checking whether under-wired code was interrupted work before deleting. It was not: 3 commits total, introduced 2026-01-05 in a bulk `feat: implement Phase 2-4 infrastructure components` drop, with only mechanical touches since (determinism migration, clamp consolidation). Speculative build-ahead.

## Not a breaking change

Neither symbol reached the public entrypoint. `dist/index.d.ts` contains **0** references to `WorkBalancer` — and as a control, the same grep returns `CompositeRouter` 28 and `ConsensusEngine` 14, so the check is meaningful rather than vacuously passing. Reachability was via the internal `context/` barrel only, not `exports/context.ts`.

Changeset is `minor`.

## What's removed

| Item | Note |
| --- | --- |
| `context/work-balancer.ts` (388 lines) | zero production consumers |
| `context/work-balancer-types.ts` | " |
| `context/work-balancer.test.ts` (42 tests) | capacity assertions ported to #4373 first |
| `context/index.ts` barrel exports | replaced with a signpost comment |
| `CompositeRouter.getCapacityDashboard()` + interface decl | no call sites; two test mocks only |
| `composite-router-helpers.fetchCapacityData` | orphaned by the above; its only caller |
| two test mocks | `delegate-to-model.test.ts`, `feedback-integration.test.ts` |

## Guarding against collateral deletion

`TaskProfile`, `CapacityInfo`, `QueuedTask` and `ScoreBreakdown` are **independent surviving types that merely shared these names** — they live in `core/task-analysis/task-profile-adapter.ts`, `adapters/capacity-monitor-types.ts`, `workflows/task-queue.ts` and `agents/experts/expert-selector-types.ts`. None was touched. A signpost comment in `context/index.ts` records this so a future reader chasing a "missing" symbol doesn't conclude it was deleted here.

## Verification

- `tsc --noEmit` — **clean**. This is the substantive proof: a green typecheck after removing the barrel export is what demonstrates production consumers were zero.
- `vitest` on the three affected suites — **165 passed**.
- `eslint` on all changed files — **clean**.

## Deliberately left stale

`docs/api/cli-adapters.md` still documents `getCapacityDashboard`. Regenerating TypeDoc rewrites ~40,000 lines across 20 files, of which only **5** relate to this change, because every permalink embeds the commit SHA. Attaching that to this PR would make it unreviewable. Filed as #4449 with measurements; the `TypeDoc Verification` gate is non-blocking by design (#2027).

`docs/architecture/ROUTING_SYSTEM.md` **is** hand-updated here — it documented `IWorkBalancer` as a live component for months while it had no consumers, so the section is rewritten to record the removal and point at #4373.


williamzujkowski:
## Correction: this IS a breaking change. Changeset bumped `minor` → `major`.

An adversarial review pass against the actual diff caught a factual error in my own PR description and changeset. Recording it plainly.

**What I got wrong.** I grepped `dist/index.d.ts` for `WorkBalancer`, got `0`, and generalised that to *both* removed symbols — I never grepped the second one. The control-symbol check I ran (`CompositeRouter=28`, `ConsensusEngine=14`) confirmed the *method* was sound while the *conclusion* was overgeneralised, which is a tidy reminder that a valid technique proves nothing about a claim it was never applied to.

**The fact.** `getCapacityDashboard` was in the public surface:

- `dist/index.d.ts:4824` — `ICompositeRouter` interface member
- `dist/index.d.ts:4999` — `CompositeRouter` class method
- `ICompositeRouter` is exported at `src/exports/cli-adapters.ts:71`, surfaced by `src/index.ts:53`

Confirmed by rebuilding on this branch — the method appeared **twice** in `dist/index.d.ts` before and **zero** times after, with `CompositeRouter` steady at 28 as a control. Removing a method from a publicly-exported interface *and* its exported implementing class breaks external callers and external implementors.

Half of my original claim does survive: `WorkBalancer` genuinely was internal-only — `exports/context.ts` re-exports only the TokenCounter subset from `context/index.ts`. The changeset now separates the two symbols instead of asserting one blanket status.

**Governance.** Breaking API changes require a *unanimous* threshold. The #4378 vote was 7/0 unanimous under `higher_order`, so the bar is formally met. But I should flag that the breaking-change dimension was never in front of the voters — my proposal framed the fork purely on architectural grounds and never raised public-API reachability. The panel cleared the strictest threshold without being told the strictest threshold applied. Reviewers may reasonably want that re-confirmed rather than inferred.

**Second fix.** The `context/index.ts` signpost was wrong about `QueuedTask`: `workflows/task-queue.ts:16` is `interface QueuedTask<T>` — module-private (no `export`) and generic — so it is neither reachable nor substitutable for the deleted non-generic type. Of the four names, only `ScoreBreakdown` is publicly re-exported; `TaskProfile` and `CapacityInfo` are module-level exports absent from every `exports/*.ts`. The comment now states exact per-symbol status.

**Deletion mechanics were confirmed clean** — no dangling references, no orphaned imports (the `CapacityStatus, ICliAdapter` import went with its only consumer; `RoutingArmId` is correctly retained at 20+ other sites), and `ITaskRouter` never declared the method, refuting that hypothesis.

Verification after the correction: full local suite **27,826 passed / 16 skipped / 0 failed**; CI **41 pass, 3 skipping, 0 failures**.

---

### A separate finding this produced: #4451

I also ran the substrate's own `pr_review` panel on this PR and — by mistake — passed a **prose summary** as `prDiff` instead of the unified diff. It returned 5/0 approve, `verified: true`, and wrote that to `governance/pr-review-records.jsonl`.

It approved a PR that genuinely warranted `request_changes`, because it never saw the code. And nothing in the record distinguishes it from a real review — `reviewedDiffHash` was the hash of my prose. Root cause: `PrReviewInputSchema` validates `prDiff` only by length; its own tests assert `prDiff: 'y'` is valid input.

That record would have been `sequence: 0`, the **inaugural entry** of the pr-review governance ledger (the committed file is empty). I dropped it rather than committing it — it was never in the append-only chain, so nothing forks, and seeding the #3849 evidence corpus with a wrong `verified: true` approve is a bad first entry. Filed as **#4451** with the full disposition.

williamzujkowski:
## Breaking-change vote: **6–1 for Option A (ship as major now)**

Live 7-voter panel, `unanimous` strategy, `errorPolicy: absolute_quorum`, no simulated votes, record persisted.

**Read the tally carefully.** The engine recorded `decision: approved, voteCounts: {approve: 7, reject: 0}`. That counts approve-vs-reject *on the proposal*, not agreement on which option. On the actual three-way question the split was:

| Option | Voters |
|---|---|
| **A — ship as major now** | architect, devex, ai_ml, pm, catfish, scope_steward (**6**) |
| C — split the PR, defer the interface removal | security (**1**) |
| B — deprecated shim | **0** |

So this is **6–1, not unanimous**, and I am recording it that way rather than quoting the 7/0 the tool emitted. (That gap between the recorded tally and the real disagreement is itself a governance-record fidelity problem — filed separately as #4452.)

### Why A won

Every A-voter converged on the same point about **Option B**: a deprecated shim could *never* be redirected to #4373's replacement, because the shapes are deliberately incompatible — an exclusion predicate is not a dashboard getter. So the "deprecation" would have no successor to point at and no removal-forcing date, making it a permanent zombie method that contradicts the #4378 decision that this surface should not exist. B got zero votes.

On **Option C**, the majority argued "hold for the next intentional major" defers to an event with no schedule: the package is at 2.178.1 on a continuous changesets cadence with no major-release train, so C is deferral to a trigger that does not exist — the exact dependency-blocked-work pattern the repo's track-deferred-work rule exists to prevent.

The shared reframe: **majors are a signaling mechanism, not a scarce budget.** The "spending the version bump cheaply" tension in my proposal rests on treating 3.0.0 as costly. For a package publishing continuously at 2.178.x, the expensive thing is unconsumed public surface, not the integer.

### The dissent is worth preserving

Security voted C on a distinct argument the majority did not rebut: *cheap majors train downstream consumers to rubber-stamp major-version upgrade reviews, eroding the attention a genuinely dangerous major depends on.* That is a supply-chain hygiene concern, not a versioning aesthetic, and it is the one argument here that gets stronger the more often the project does this. Worth revisiting if a pattern of low-content majors develops.

Security also agreed B was the worst option — the dissent was A-vs-C on sequencing, not a defence of the shim.

### Binding condition (architect, already satisfied)

> the changeset/release notes must name the removed method, state that `ICliAdapter.getCapacity()` remains as the data source, and point to #4373 as the successor capability

The committed changeset does all three. Verified before merge.

### Process note

This vote exists because the original #4378 proposal never mentioned public-API reachability, so the 7/0 panel cleared the unanimous breaking-change bar without being told the bar applied. Several voters independently flagged that this corrective vote is the right remedy for that disclosure defect. The architectural outcome is unchanged; what changed is that it is now decided with the relevant fact in evidence.

Issue #4351, opened 2026-08-03
bug(mcp): fail closed when adapters exhaust capacity or orchestration produces no result


williamzujkowski:
## Description

The MCP orchestration surfaces can report a successful/complete outer job even when no model work completed, no consensus result exists, and the nested execution explicitly says `completed: false`.

In a second live call, a seven-role security/storage consensus panel returned `stop_sequence` for every voter with zero tokens and no decision. The installed health checks still report three authenticated CLI adapters and a ready CLI-subprocess voter transport.

The operator confirmed that the Claude Code and AGY adapters had exhausted their weekly context/token allowances during this reproduction. Nexus did not represent that capacity state accurately, exclude those adapters from routing, or explain it in the terminal result.

This is distinct from, but exposed alongside:

- #4348 / #4349: Codex MCP sends the obsolete `on-failure` approval policy.
- #4350: authenticated CLI adapters are ignored by the PR-review path.

The missing contract is fail-closed behavior at the MCP orchestration boundary: an adapter/configuration failure or an empty voter panel must not be surfaced as a successful completed job.

## Steps to Reproduce

Environment and repository:

- Date: 2026-08-02
- nexus-agents: `2.173.6`
- Node.js: `v22.22.3`
- npm: `10.9.8`
- OS: Ubuntu 24.04.1, Linux 7.0.0-28-generic x86_64
- Consumer repository: `williamzujkowski/progquest`
- MCP transport: stdio through Codex

### 1. Confirm advertised health

Run:

```text
nexus-agents verify
nexus-agents doctor
```

Observed health summary:

```text
Adapter Availability: 3 authed CLI(s): claude, codex, opencode
MCP Server mode: Ready
MCP Client mode: Ready (Codex mcp-server)
Voter transport: CLI subprocess
```

`doctor` also reported Claude CLI capacity as `100% remaining`, despite the operator-observed weekly context/token exhaustion. Authentication therefore passed while usable capacity did not.

### 2. Run a live research orchestration through the MCP `run` tool

Payload:

```json
{
  "goal": "Research and evaluate a fail-closed browser save and clipboard recovery contract for ProgQuest, including typed outcomes, explicit user actions, accessibility, and security trade-offs.",
  "execute": true,
  "forceStrategy": "research",
  "dispatch": "async",
  "dependencyStructure": "linear",
  "isNovel": false,
  "requiresConsensus": true
}
```

Poll the returned job until terminal.

Reproduced twice:

[...]

Issue #4303, opened 2026-07-18
harden(resilience): classifyExpertFailure can reject Promise<Result> on an Error with a throwing .message/.cause getter (defensive try-guard)


williamzujkowski:
## Context
Flagged (out-of-scope, non-blocking) by the security lens on BOTH the #4292 and #4302 vote-gates. Pre-existing since #4286 — not introduced by #4293.

## Problem
`RecoverableExpert.execute()` (`packages/nexus-agents/src/agents/experts/expert-recovery.ts`) runs `classifyExpertFailure` inside `withRetry`'s `isRetryable` predicate — and `withRetry`'s catch block is not itself try-guarded (`adapters/retry.ts`). `classifyExpertFailure` walks the error's `.cause` chain and reads `.message` (via `getErrorMessage`/`extractErrorMessage`). An `Error`-like object with a **throwing `.message` or `.cause` getter** would throw out of the predicate, escaping `withRetry` and **rejecting the `Promise<Result<...>>`** instead of returning `err(...)` — breaking the never-throws contract that callers (`execute_expert`) rely on. The same applies to the new `annotateExhausted` re-classification path (#4293).

## Realism / severity
LOW — expert-execution errors originate from the model/adapter/agent layer (`AgentError`, `ModelError`, HTTP-status errors), not attacker-supplied objects, and the `MAX_CAUSE_DEPTH`/seen-set guards already bound traversal. But it's a real fail-closed-contract gap.

## Fix
Wrap the `isRetryable` classification predicate (and the `annotateExhausted` re-classification) in a defensive try/catch that treats a classifier throw as `permanent` (fail closed) — or harden `getErrorMessage`/`extractErrorMessage` to swallow getter throws. Add a test with an `Error` whose `.message` getter throws → `execute()` returns `err(...)`, never rejects.

## Trigger
Do now as cheap hardening, or when expert execution can ever receive externally-constructed error objects. Refs #4286, #4293, PRs #4292/#4302.

_Filed autonomously — surfaced twice by the resilience vote-gates._

williamzujkowski:
Already done — verified during backlog grooming (2026-08-14).

`classifyExpertFailure` is wrapped exactly as this issue asked (`agents/experts/expert-recovery.ts:204-240`):

```ts
  } catch {
    // A throwing `.message`/`.cause` getter (or any classifier fault) MUST NOT
    // escape — fail closed so `execute()` still resolves to a Result.
[...]

Issue #807, opened 2026-02-06
feat: Centralized Model Registry + Quality-First Routing


williamzujkowski:
## Summary

Centralize all model metadata into the canonical registry (`config/model-capabilities.ts`) and implement quality-first routing defaults.

## Problem

Model metadata (pricing, display names, context windows, max output, quality scores, CLI aliases, defaults) is duplicated across 10+ files. A single model change requires editing 3-5 files.

### Change 0: Centralize Model Registry
- **0a**: Extend `ModelCapabilitySchema` with `pricing`, `qualityScores`, `maxOutputTokens`, `cliName`, `cliAlias`, `cliModelName`
- **0b**: Add data to each of 8 model entries + `DEFAULT_MODEL_PER_CLI` constant
- **0c**: Create `config/model-config-helpers.ts` with derived helper functions
- **0d**: Replace duplicated sources across codebase:
  - `core/trace-pricing.ts` — canonical model entries
  - `cli-adapters/adapters/claude-adapter.ts` — `MODEL_TO_CLI_ALIAS`, display names, costs
  - `cli-adapters/adapters/gemini-adapter-helpers.ts` — display names, context windows, costs
  - `mcp/tools/delegate-to-model-types.ts` — `MODEL_CAPABILITIES`
  - `cli-adapters/types-capability.ts` — `DEFAULT_CAPABILITIES`
  - `cli-adapters/topsis-types.ts` — `DEFAULT_MODEL_PROFILES`
  - `testing/adapters/mock-adapter-helpers.ts` — `MODEL_INFO_BY_NAME`

### Change 1: Quality-First Model Defaults
- `DEFAULT_MODEL_PER_CLI` uses strongest model per provider (opus, gemini-pro, codex-5.3)

### Change 2: Exploration Task → Gemini Routing
- Add `EXPLORATION_KEYWORDS` and `needsExploration` to `TaskRequirements`
- Scoring bonus for large context window models on exploration tasks

### Change 3: Capacity-Aware Load Balancing
- `fetchCapacityData()` and `orderByCapacity()` in `composite-router-helpers.ts`
- Deprioritize exhausted CLIs in routing

### Change 4: Capacity Dashboard
- `getCapacityDashboard()` method on `CompositeRouter`

## Verification
```bash
pnpm typecheck
pnpm vitest run packages/nexus-agents/src/config/
pnpm vitest run packages/nexus-agents/src/cli-adapters/
[...]

PR #4454, opened 2026-08-15
feat(cli-adapters): capacity-aware routing exclusion (#4373)

### LINKED ISSUE #4373: feat(cli-adapters): normalized capacity_exhausted diagnostic + capacity-aware routing exclusion (#4351 criterion 3) — exclusion held on #4456

williamzujkowski:
Closes #4373 — criterion 3 of open bug #4351. PR 2 of 2 in the sequence set by the #4378 vote; #4450 (the deletion) landed first.

## What it does

`CapacityFilterStage` excludes a routing candidate whose adapter reports **measurably** exhausted capacity, with a normalized `capacity_exhausted` diagnostic. It runs with the other hard filters, before any scoring — no point scoring an arm that cannot serve.

This restores the capacity semantics of the `WorkBalancer` deleted in #4450, in the shape the router actually needs. Only the predicate carried over; the queue was the shape mismatch that decided that vote. Its capacity assertions are this stage's seed tests, per the vote's binding condition.

## The design constraint that shaped it

`CapacityStatus` carries an `observed` flag (#4374), and its docstring is explicit: when false, **every other field is a default, not a measurement** — an untracked adapter reports its full token limit and 0% utilization, indistinguishable from a genuinely idle one.

This invalidated the naive port of `WorkBalancer`'s `remainingTokens / estimatedTokens` formula I had originally planned: against an unobserved adapter it computes a ratio from a default and scores it maximally healthy from zero data. That is the same absence-coerced-into-a-value defect as #4435/#4439.

So candidates classify three ways, and the third is the point:

| State | Effect |
| --- | --- |
| `exhausted` — `observed && (exhausted \|\| remainingTokens <= 0)` | Excluded |
| `healthy` — observed, capacity remaining | Kept |
| `unmeasured` — not observed, no adapter, or probe failed | Kept, counted separately |

- **Unmeasured never excludes** — fails open on absent evidence, because exclusion is destructive.
- **Unmeasured is never counted as healthy either** — it surfaces as a distinct `capacity:unmeasured-N` signal. Collapsing the two is the blindness #4436 was filed about.

## Default posture: enforce (7/7 vote, unanimous on the option)

Full record on #4373. The panel's argument was better than the one I put in the proposal: the tracker sees only *this* process's spend, so `remainingTokens` is an **upper bound** — an observed exhaustion implies provider-side quota is at least as exhausted. The residual error is therefore all on the false-negative side, which is the pre-existing bug. Warn-mode was rejected outright because nothing consumes its telemetry, so it degenerates into off-by-default (#2795 and #4262 cited as this repo's own evidence).

**Known limitation, stated plainly:** `observed` guards against *absent* evidence, not *wrong* evidence — a stale or misconfigured local limit still yields a confident false positive. And quota burned by another process is invisible. **This closes criterion 3 for locally-observable exhaustion only**, not #4351's full scenario.

## Reusing a flag that had been lying since #807

Wired to the existing `enableCapacityBalancing` rather than adding `enableCapacityFilter`. That flag has defaulted to `true` since #807 while promising "deprioritize exhausted CLIs" — and **no stage ever read it**. It appears in `STAGE_FLAG_KEYS` for config passthrough and nowhere else. Reusing it makes it honest instead of creating a second switch for one concern, which is the mistake #4378 was just deleted for. Wording changed from "deprioritize" to "exclude"; nothing can depend on the former, since it never existed.

I also caught myself reaching for a non-existent `armToSlot` and switched to the canonical `keepArmsForSlots` (#3422), which preserves distinct `api:*` arms rather than collapsing them.

## A resilience bug the full suite caught

Six `pipeline-e2e` tests failed after wiring, with the **whole routing call** rejecting. That suite's mock adapter omits `getCapacity`, and calling a missing method throws a `TypeError` **synchronously inside `.map()`** — before `Promise.allSettled` receives anything. My fail-open handled rejected probes but not synchronous throws.

Consequence beyond the fixture: a structurally partial adapter would take down routing for **every** candidate. A capacity filter causing total routing failure is the exact inverse of its purpose, and the same never-throws class as #4303/#4308. Fixed with an `async` callback so the throw becomes a rejection.

**Verified by mutation, not by going green:** reverting the fix fails both the e2e suite *and* the new dedicated regression test together; restoring turns both green. My 14 stage tests stayed green throughout, because every one supplies a well-formed stub — a targeted test file cannot find a defect that only appears against a differently-shaped mock.

## Binding condition from the vote, implemented

Two voters required the all-excluded path be explicit and diagnosed. My first cut raised a bare code; it now names every excluded arm and reason, with a test asserting each name appears. Leaving it bare would have reproduced #4351's actual complaint that nexus "did not explain it in the terminal result".

## Verification

- Full suite: **27,845 passed / 16 skipped / 0 failed**
- `tsc --noEmit` clean; `eslint` clean; `markdownlint` clean
- 18 new tests (14 stage-level + 4 runner-level)
- Docs: added a Capacity Filter Stage section **and** added it to the Overview executed-stage diagram — that diagram's own caption warns an outdated version "gives the wrong mental model when debugging 'why was my model rejected?'" (#2947), and capacity exclusion is precisely that answer

### DIFF
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -4,14 +4,16 @@
 
 Add capacity-aware routing exclusion (#4373, criterion 3 of #4351).
 
-`CapacityFilterStage` removes a routing candidate whose adapter reports **measurably** exhausted capacity, emitting a normalized `capacity_exhausted` diagnostic. When every candidate is excluded the router now fails closed with an error naming each excluded arm and why, instead of routing to an adapter that cannot serve.
+`CapacityFilterStage` classifies each routing candidate's adapter capacity and emits a normalized `capacity_exhausted` diagnostic. Under `enforceHardLimits: true` it removes exhausted candidates, and when every candidate is excluded the router fails closed with an error naming each excluded arm and why. **The shipped default is signal-only** — see below.
 
 This restores — in the shape the router actually needs — the capacity semantics of the `WorkBalancer` removed in #4378. Only the predicate carried over; the queue did not.
 
 **Unmeasured capacity never excludes.** `CapacityStatus.observed` (#4374) marks whether a reading is real: when false, every other field is a default rather than a measurement. The stage classifies each candidate as exhausted / healthy / **unmeasured** and only an observed reading can exclude. A missing adapter or a failed capacity probe is likewise unmeasured, never exhausted — it fails open on absent evidence. Unmeasured candidates are also not counted as healthy; they surface as a distinct `capacity:unmeasured-N` signal.
 
-Enforcing by default (7/7 consensus vote). The local tracker sees only this process's spend, so `remainingTokens` is an upper bound — meaning an _observed_ exhaustion implies provider-side quota is at least as exhausted. The residual error is therefore false negatives, which is the pre-existing behaviour this fixes.
+**Signal-only by default — it does not exclude anything unless you opt in.** The only capacity signal available today is not quota exhaustion: `CapacityTracker` sets `exhausted` from a rolling 60-second window against hardcoded per-minute estimates (claude 100k tokens / 50 requests, its own comment calls them "conservative estimates"), and it self-clears within the minute. Hard-excluding on that would let an ordinary burst — a 7-voter panel, a subagent fan-out — empty the candidate pool and fail routing closed for a condition that resolves itself. Pass `enforceHardLimits: true` if you have a signal you trust.
 
-**Known limitation:** the tracker has no visibility into provider-side quota consumed by other processes, so exhaustion burned elsewhere is still invisible. This closes criterion 3 for locally-observable exhaustion only.
+**This does NOT yet close #4351 criterion 3.** That bug was weekly _quota_ exhaustion, which a per-minute rate counter cannot detect. #4456 tracks adding a durable, provider-asserted quota signal; enforcement is held until then.
 
-Wired to the existing `enableCapacityBalancing` config flag rather than a new one. That flag was declared for #807 and defaulted to `true` while promising "deprioritize exhausted CLIs", but no stage ever read it — it is now honest. Set it to `false` to disable, or construct the stage with `enforceHardLimits: false` for signal-only operation.
+**Second limitation (#4455):** capacity is assessed per display slot, so a CLI arm and an `api:*` arm sharing a slot share a verdict. Not reachable under the default `plan` billing mode.
+
+Wired to the existing `enableCapacityBalancing` config flag rather than a new one. That flag was declared for #807 and defaulted to `true` while promising "deprioritize exhausted CLIs", but no stage ever read it — it is now honest. Set it to `false` to disable the stage entirely.
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,17 @@
+---
+'nexus-agents': minor
+---
+
+Add capacity-aware routing exclusion (#4373, criterion 3 of #4351).
+
+`CapacityFilterStage` removes a routing candidate whose adapter reports **measurably** exhausted capacity, emitting a normalized `capacity_exhausted` diagnostic. When every candidate is excluded the router now fails closed with an error naming each excluded arm and why, instead of routing to an adapter that cannot serve.
+
+This restores — in the shape the router actually needs — the capacity semantics of the `WorkBalancer` removed in #4378. Only the predicate carried over; the queue did not.
+
+**Unmeasured capacity never excludes.** `CapacityStatus.observed` (#4374) marks whether a reading is real: when false, every other field is a default rather than a measurement. The stage classifies each candidate as exhausted / healthy / **unmeasured** and only an observed reading can exclude. A missing adapter or a failed capacity probe is likewise unmeasured, never exhausted — it fails open on absent evidence. Unmeasured candidates are also not counted as healthy; they surface as a distinct `capacity:unmeasured-N` signal.
+
+Enforcing by default (7/7 consensus vote). The local tracker sees only this process's spend, so `remainingTokens` is an upper bound — meaning an _observed_ exhaustion implies provider-side quota is at least as exhausted. The residual error is therefore false negatives, which is the pre-existing behaviour this fixes.
+
+**Known limitation:** the tracker has no visibility into provider-side quota consumed by other processes, so exhaustion burned elsewhere is still invisible. This closes criterion 3 for locally-observable exhaustion only.
+
+Wired to the existing `enableCapacityBalancing` config flag rather than a new one. That flag was declared for #807 and defaulted to `true` while promising "deprioritize exhausted CLIs", but no stage ever read it — it is now honest. Set it to `false` to disable, or construct the stage with `enforceHardLimits: false` for signal-only operation.
--- a/docs/architecture/ROUTING_SYSTEM.md
+++ b/docs/architecture/ROUTING_SYSTEM.md
@@ -326,9 +326,40 @@ The decisive problem was not that it was unused, but what it contained: alongsid
 
 `CompositeRouter.getCapacityDashboard()` and its helper `fetchCapacityData` were removed in the same change — a read-only surface whose only references were two test mocks.
 
-**Capacity-aware routing is not gone; it is relocating.** #4373 implements capacity exclusion as a predicate inside the existing stage chain (`Budget → Zero → Preference → Topsis → LinUCB`), which is the shape the routing path actually needs — a per-candidate decision input rather than a queue. The deleted component's capacity-semantics tests (exhausted flag, zero-remaining, the estimate-vs-remaining boundary) are preserved on #4373 as its seed specification.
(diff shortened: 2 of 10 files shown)


williamzujkowski:
## Self-review finding before merge: #4455 (documented, not fixed here)

Tracing the highest-risk area of this change — the slot↔arm mapping — turned up a real defect. Filing and documenting rather than silently inheriting it.

**Capacity is assessed per display *slot*, but quota belongs to a serving *route*.**

`runCapacityStage` builds its context with `armsToSlots`, which de-duplicates `api:*` arms onto their vendor slot (`cli-adapters/types-core.ts:61-74`): `api:anthropic → claude`, `api:openai → codex`, etc. So with candidates `['claude', 'api:anthropic']` the stage probes only the **CLI** adapter and then applies that verdict to both arms via `keepArmsForSlots`.

Two observable failures:

- **False positive (the destructive one).** CLI `claude` observed-exhausted, `api:anthropic` healthy → both arms removed. A serviceable API arm with its own independent quota is excluded because a *different* serving route ran out. If it was the last candidate, the task fails closed naming an arm that had capacity.
- **False negative.** CLI healthy, api arm exhausted → nothing filtered → the router may select the exhausted arm, which is the exact #4351 failure this stage exists to prevent.

**Reachability:** not under the default configuration. API arms register only under `NEXUS_BILLING_MODE=api` (`factory.ts:127-131`, citing #3422) and `billingMode` defaults to `'plan'`. Reachable when that env var is set.

**Why I'm not fixing it in this PR.** Slot granularity is pre-existing and shared by every `RoutingContext`-based stage — `armsToSlots`' own docstring documents the de-duplication. For *scoring* stages it is a fair approximation. Capacity is simply the first stage where the shared abstraction yields the **wrong quantity** rather than an imprecise one, because quota is per-credential and the action is exclusion rather than a score nudge. Fixing it properly means assessing per-arm outside the slot abstraction, which is #4391's gateway work (separate model identity from serving route) — the same separation #4392 is about.

Documented in two places so it cannot be inherited silently: the `runCapacityStage` docstring and the ROUTING_SYSTEM.md capacity section. Filed as **#4455** with a suggested interim guard (treat a multi-arm slot as `unmeasured`, which fails open and removes the destructive false positive) should api billing mode see use before #4391 lands.

Also note for reviewers: I have **not** run this PR through `pr_review`. That tool accepts arbitrary text as `prDiff` and would write a `verified: true` governance record I could not vouch for (#4451). Review here is a `code-reviewer` agent reading the actual `git diff main...HEAD`, plus this self-review.

williamzujkowski:
## Correction: the 7/7 enforce-by-default vote rested on a false premise I supplied. Default flipped to signal-only.

Adversarial review of the actual diff found that `CapacityStatus.exhausted` does not mean what I told the panel it meant. The vote's decisive reasoning does not survive the correction, so I am not shipping its conclusion.

### What I told the panel

> The local tracker sees only THIS process's spend, so `remainingTokens` is a local upper bound, never authoritative. Consequence: the stage can MISS a genuinely exhausted adapter (false negative), but a false POSITIVE requires the local tracker to have actually observed exhaustion.

Every voter built on that. The architect: *"the known error mode is the false NEGATIVE… so enforcing cannot regress availability relative to the status quo."* The scope steward: *"if the tracker OBSERVED exhaustion, the provider-side quota is at least as exhausted. An observed-exhaustion exclusion is close to tautologically correct."*

### What it actually is

`cli-adapters/capacity-tracker.ts`:

```ts
export const DEFAULT_TOKEN_LIMITS   = { claude: 100_000, ... };  // "conservative estimates"
export const DEFAULT_REQUEST_LIMITS = { claude: 50, ... };
export const RATE_LIMIT_WINDOW_MS   = 60_000;
// :184
const exhausted = remainingTokens === 0 || remainingRequests === 0;
```

`exhausted === true` means **this process made 50 claude calls, or spent 100k claude tokens, in a rolling 60 seconds** — against a hardcoded guess at one API tier — and it self-clears within the minute (`resetTime = windowStart + windowMs`). `recordUsage` (`:151`) pushes a request timestamp unconditionally, so request-count exhaustion trips even with no token accounting at all. It is live from the first call: `base-adapter.ts:213` feeds it after every execution and `hasObserved` flips true immediately.

### Why the vote's logic fails

"Local spend is a lower bound on provider-side spend" is true of *quota*. It is not true of a *per-minute rate window*. Burning 50 requests in 60 seconds implies nothing whatsoever about weekly plan quota — the two are different quantities, not the same quantity measured incompletely. So:

- **The premise that false positives are near-impossible is wrong.** A 7-voter consensus panel or a subagent fan-out is exactly the traffic shape that trips a 50-req/60s counter while plenty of quota remains.
- **The claimed fix does not fix the reported bug.** #4351 was *weekly quota* exhaustion. Quota burned gradually, or burned by another process, never trips a rolling-60s counter. The stage as voted could not have detected the incident it was built for.
- **Enforcing would add a failure mode.** Once all four slots trip, `runCapacityStage` returns an error and routing hard-fails — with no retry or backoff — for a condition that clears itself in under a minute. A self-inflicted outage.

### What I changed

**`DEFAULT_CONFIG.enforceHardLimits` is now `false`.** The stage classifies, traces, and emits `capacity_exhausted` and `capacity:unmeasured-N` signals, but does not remove candidates. `enforceHardLimits: true` remains available for a caller with a trustworthy signal. A test at the *wiring* boundary asserts the shipped default is signal-only, so re-enabling enforcement must break a test rather than pass silently.

Filed **#4456** for the actual gap: there is no durable quota signal anywhere in the tree. It proposes distinguishing `rateLimited` (short, self-clearing) from `quotaExhausted` (durable, provider-asserted, from rate-limit headers or a classified quota error per #4382), and enforcing only on the latter.

### On the option the panel rejected

The panel dismissed warn-only 7/0, largely on my framing that nothing consumes warn telemetry so it degenerates into off-by-default. That objection was reasonable given what I told them, and it is now answered on its own terms: the named consumer of the signal is **#4456's decision about whether the signal is trustworthy enough to enforce on** — which is a real question with a real owner, not an indefinite park.

I am not re-running the vote. The panel decided a question whose factual basis I got wrong; re-asking it with corrected facts would be reasonable, but the corrected facts make signal-only the only safe option, and there is no fork left to put to a panel. **#4373 therefore does not close #4351 criterion 3** — I will leave that criterion open on #4351 rather than mark it done.

### Also fixed from the same review

- **Unguarded boundary throw** — `runCapacityStage` awaited `route()` without a try/catch, so an escaped throw would reject the whole routing call. That is the same failure class as the synchronous-throw bug I had already fixed *inside* `assessAll`; guarding one level down while leaving the boundary open only moved it.
- **No probe timeout** — `getCapacity()` is an unbounded promise on a public interface and `maxDecisionTimeMs` is declared but enforced nowhere, so one hanging adapter stalled all routing. Added `probeTimeoutMs` (2s); a timed-out probe is `unmeasured`, never `exhausted`.
- **A tautological test** — `expect(signals).not.toContain('capacity:unmeasured-0')` sat inside a case where the count was 1, so no implementation could fail it. Replaced with a real complement asserting no unmeasured signal when every candidate was measured.
- **JSDoc contradicting code** — a comment claimed the unmeasured signal was emitted at zero to distinguish "all healthy" from "all unknown"; it never was. Corrected to describe actual behaviour.

Issue #4452, opened 2026-08-15
governance(consensus_vote): multi-option proposals record as 'unanimous 100%' while masking a real split — a 6-1 vote was persisted as 7-0


williamzujkowski:
## Summary

`consensus_vote` records an approve/reject/abstain tally. When a proposal asks voters to choose among named options (A/B/C), every voter who engages constructively returns `approve` — so the persisted record shows **unanimous approval** even when the panel substantively disagreed about *which option*. The recorded verdict can therefore overstate consensus on the question actually being decided.

## Concrete instance (2026-08-15)

I ran a `unanimous`-strategy vote on PR #4450 asking voters to choose A, B, or C for how to handle a breaking API removal. The persisted record:

```json
{"strategy":"unanimous","decision":"approved","approvalPercentage":100,
 "voteCounts":{"approve":7,"reject":0,"abstain":0,"error":0}}
```

The actual distribution, recoverable only by reading seven free-text `reasoning` fields:

| Option | Voters |
| --- | --- |
| A — ship as major now | 6 (architect, devex, ai_ml, pm, catfish, scope_steward) |
| C — split the PR | 1 (security) |
| B — deprecated shim | 0 |

**6–1, recorded as 7–0 unanimous.** Nothing in the structured record indicates a dissent existed. Anyone consuming the record downstream — a promotion gate, an audit review, a future agent — reads unanimity.

The dissent was substantive, not a quibble: security argued that routinely spending major bumps on low-content breaking changes trains downstream consumers to rubber-stamp major-version upgrade reviews, degrading a supply-chain safety signal. That is exactly the kind of minority view a governance record exists to preserve.

## Why this matters

1. **Threshold semantics get inverted.** The strictest thresholds (`unanimous` for breaking API changes, `supermajority` for architecture/security) are meant to be *hard* to clear. On a multi-option proposal they become trivially easy: everyone approves, nobody is choosing the same thing. A 4/3 split across two options would still record as `unanimous, 100%`.
2. **The record is the decision.** Per repo convention the vote result *is* the outcome, and vote records feed the authority-ladder ratification path (`ratifies`) and the audit chain. A record asserting unanimity where a 6–1 split occurred is misleading evidence, and it is misleading in the direction of overconfidence.
3. It is only recoverable by prose archaeology. I extracted the real split by reading every `reasoning` string. No structured consumer can do that.

## Relationship to other work

[...]

PR #4458, opened 2026-08-15
fix(mcp): reject a non-diff prDiff in pr_review (#4451)

### LINKED ISSUE #4451: governance(pr_review): prDiff accepts arbitrary prose — a summary produces a 'verified: true' approval record indistinguishable from a real diff review
## Summary

`pr_review` performs no shape validation on `prDiff`. Any non-empty string is accepted, so a **prose summary of a change** produces a governance record indistinguishable from a review of the actual diff — including `verified: true` and a `reviewedDiffHash` that implies a diff was hashed.

This was found by doing it accidentally, then catching it.

## What happened (2026-08-15, PR #4450)

I invoked `pr_review` on PR #4450 and passed a **prose description** of the change as `prDiff` instead of the unified diff. The panel returned a clean approval, and this landed in `governance/pr-review-records.jsonl`:

```json
{"version":"1.1","sequence":0,"prNumber":4450,
 "baseSha":"1f90f2afdb7e830988823e4257520c495dfc282b",
 "reviewedDiffHash":"3327f44a019a414a8e76991b04a9a0cbf0075e8d4704117c7b4a9efd8da89a44",
 "verdict":"approve","verified":true,
 "voteCounts":{"approve":5,"request_changes":0,"abstain":0,"error":0,"total":5}}
```

Nothing in that record distinguishes it from a real diff review. `reviewedDiffHash` is the hash of my prose.

**The panel's approval was wrong, and demonstrably so.** A `code-reviewer` agent given the *actual* `git diff main...HEAD` found a genuine defect the panel could not have seen: the changeset was labelled `minor` while the PR removes `getCapacityDashboard()` from the publicly-exported `ICompositeRouter` interface — a breaking change. Verified by rebuilding: `dist/index.d.ts` contained the method **twice** before the change and **zero** times after, with `CompositeRouter` steady at 28 occurrences as a control.

So the ledger holds a 5/0 `verified: true` approve for a PR that genuinely warranted `request_changes`.

## Root cause

`PrReviewInputSchema` constrains `prDiff` only by length (`minLength: 1`, `maxLength: 2000000`). The tool's own tests assert that a single character is valid input:

- `mcp/tools/pr-review-tool.test.ts:481` — `PrReviewInputSchema.safeParse({ prTitle: 'x', prDiff: 'y' })` expected to pass
- also `:456`, `:461`, `:487`, `:492`, `:494`

There is no check for `diff --git`, `--- a/` / `+++ b/`, or `@@` hunk headers anywhere in the input path.

## Why this matters more than an ordinary input-validation gap

The pr-review ledger is the evidence base for the `#3849` audit→enforce promotion and epic `#4226` (feeding the governance ledger from all live diff-bound reviews). Its value depends entirely on records being what they claim. A record whose `reviewedDiffHash` hashes prose is worse than a missing record, because it is counted as evidence.

It also sits squarely in `#3829` (governance-of-the-governor): a caller can obtain a `verified: true` approval without a real diff ever reaching a voter, with no signal at the record level.

Related: `#3927` (enforce + harden authentic vote records) is the vote-side analogue of this problem.

## Proposed fix

1. **Validate `prDiff` shape at the schema boundary.** Require at least one recognizable unified-diff marker (`diff --git`, a `--- `/`+++ ` file-header pair, or an `@@ ... @@` hunk header). Fail closed with a clear validation error, consistent with the untrusted-input "fail closed on ambiguity" rule.
2. **Record the provenance in the audit record.** Add a field distinguishing a validated unified diff from unvalidated free text, so historical records can be filtered rather than silently trusted.
3. **Do not retroactively delete the bad record.** The chain is append-only and tamper-evident; deleting is worse than annotating. Prefer a follow-up record or an explicit provenance field.
4. Update the tool description to state that `prDiff` must be a unified diff — it currently says "Unified diff text" but nothing enforces it.

## Acceptance

- [ ] `PrReviewInputSchema` rejects prose; the existing tests asserting `prDiff: 'y'` is valid are updated to assert rejection
- [ ] A rejected call writes **no** governance record
- [ ] Records carry a diff-provenance indicator
- [ ] Existing record for PR #4450 (`sequence: 0`) is annotated, not removed

Found while reviewing #4378 / PR #4450.

williamzujkowski:
Closes #4451.

## The defect

`prDiff` was validated by **length alone**. Its own tests asserted `prDiff: 'y'` was valid input. So a prose summary of a change was accepted, ran the full 5-voter panel, and persisted a governance record:

```json
{"prNumber":4450,"reviewedDiffHash":"3327f44a…","verdict":"approve","verified":true,
 "voteCounts":{"approve":5,"request_changes":0,"abstain":0,"total":5}}
```

Nothing in that record distinguishes it from a review of real code. `reviewedDiffHash` was the hash of my prose.

**And the verdict was wrong.** A `code-reviewer` agent given the *actual* `git diff` found a defect the panel could not have seen — PR #4450's changeset was labelled `minor` while removing `getCapacityDashboard()` from the publicly-exported `ICompositeRouter`. The panel approved a PR that warranted `request_changes`, because it never saw any code.

I found this by doing it accidentally, then catching it. That record would have been `sequence: 0` — the **inaugural entry** of the pr-review ledger, which is the evidence base for the #3849 audit→enforce promotion. I dropped it rather than committing it (it was never in the append-only chain, so nothing forks).

## The fix

A line-anchored shape gate at the schema boundary. Accepts **any** unified diff, not only git's:

| Marker | Source |
| --- | --- |
| `^diff --git ` | git |
| `^--- ` / `^+++ ` | `diff -u`, `svn diff`, patch files |
| `^@@ … @@` | bare hunk |

Rejects only when none appear. All anchored under `/m`, so prose that merely *mentions* the tokens does not qualify — `'I ran diff --git and saw @@ markers'` is rejected, and there's a test for it.

**Deliberately not git-only.** This file's own fixture at `:374` is `'--- a/foo.ts\n+++ b/foo.ts\n@@ -1 +1 @@\n-old\n+new'` — a legitimate diff with no `diff --git` line. A narrower gate would have rejected one of the tool's own test cases, which is a useful signal that the accept set is right.

## Reusing what exists

`looksLikeUnifiedDiff` lives beside `splitByFile` in `pr-review-diff-budget.ts`, the one module that already knows diff structure — described in its own header as "PURE, deterministic, and I/O-free".

`splitByFile` already computed the needed signal and discarded it: input with no `diff --git` header returns a single `(unstructured)` segment. **I did not change that behaviour.** Its tolerance is correct for *its* caller — the budget packer should pack whatever it is handed. The gate belongs at the entry boundary, not inside a function with a different job.

## What updating the tests revealed

Eleven call sites passed `prDiff: 'y'`, and the size tests used `'a'.repeat(50_001)`. None were testing diff handling — they cover dispatch (#3731), the simulate gate (#4170), and `repoPath` (#4278). Every one had silently relied on the missing gate.

Hoisted a shared `MINIMAL_DIFF` fixture so each exercises the field it names. The size test is now padded with diff-shaped lines so it still crosses 50k while staying structurally valid — it tests the size boundary rather than accidentally testing nothing.

Added a test reproducing the exact defect: the prose payload I actually passed, asserted rejected with a message naming "unified diff".

## Verification

- Full suite: **27,858 passed / 16 skipped / 0 failed**
- `mcp/` + `audit/` subtrees: 4,324 passed
- `tsc --noEmit` clean, `eslint` clean
- 7 new predicate tests + 2 new schema tests

Checked every other `prDiff` caller in the tree: `scripts/pr-review-local.ts`, `pr-review-local-ledger.ts` and `pr-review-eval-run.ts` all source real `git diff` / `gh diff` output, so none is affected.

## Scope — what this does not do

Closes the **input** path only. It cannot tell whether a syntactically valid diff is the diff *actually under review*, and it does not mark records already written. Record provenance metadata remains tracked on #4451's acceptance list; #4452 covers the sibling problem in `consensus_vote` (multi-option votes recording as unanimous).

### DIFF
--- a/.changeset/tidy-jars-attack.md
+++ b/.changeset/tidy-jars-attack.md
@@ -1,0 +1,13 @@
+---
+'nexus-agents': patch
+---
+
+`pr_review` now rejects a `prDiff` that is not a unified diff (#4451).
+
+Previously `prDiff` was validated by length alone, so **a prose summary of a change was accepted**: it ran a full 5-voter panel and persisted a `verified: true` governance record whose `reviewedDiffHash` was the hash of the prose. Nothing in the record distinguished it from a review of real code, and the panel approved a PR that warranted `request_changes` — because it never saw any code.
+
+Validation accepts any unified diff, not only git's: a `diff --git` header, a `---`/`+++` file-header pair, or an `@@ … @@` hunk header. All markers are line-anchored, so prose that merely mentions them does not qualify. Non-git diffs from `diff -u`, `svn diff`, or a patch file remain valid.
+
+The predicate (`looksLikeUnifiedDiff`) lives beside `splitByFile` in `pr-review-diff-budget.ts` so one module owns what a diff looks like. `splitByFile`'s own tolerance of unstructured input is unchanged — that is correct for the budget packer, which should pack whatever it is handed; the gate belongs at the entry boundary.
+
+This closes the input path. It does not retroactively mark existing records, and it cannot tell whether a syntactically valid diff is the diff actually under review — record provenance metadata remains tracked on #4451.
--- a/.changeset/tidy-jars-attack.md
+++ b/.changeset/tidy-jars-attack.md
@@ -6,8 +6,16 @@
 
 Previously `prDiff` was validated by length alone, so **a prose summary of a change was accepted**: it ran a full 5-voter panel and persisted a `verified: true` governance record whose `reviewedDiffHash` was the hash of the prose. Nothing in the record distinguished it from a review of real code, and the panel approved a PR that warranted `request_changes` — because it never saw any code.
 
-Validation accepts any unified diff, not only git's: a `diff --git` header, a `---`/`+++` file-header pair, or an `@@ … @@` hunk header. All markers are line-anchored, so prose that merely mentions them does not qualify. Non-git diffs from `diff -u`, `svn diff`, or a patch file remain valid.
+Validation accepts any unified diff, not only git's: a `diff --git` header, an `@@ … @@` hunk header, or the `---`/`+++` file-header **pair**. All markers are line-anchored, so prose that merely mentions them does not qualify. Non-git diffs from `diff -u`, `svn diff`, or a patch file remain valid, as do rename-only, binary, mode-only and CRLF diffs.
 
-The predicate (`looksLikeUnifiedDiff`) lives beside `splitByFile` in `pr-review-diff-budget.ts` so one module owns what a diff looks like. `splitByFile`'s own tolerance of unstructured input is unchanged — that is correct for the budget packer, which should pack whatever it is handed; the gate belongs at the entry boundary.
+The `---`/`+++` markers are required together rather than alone: a lone `---` line is too weak a signal, since prose uses dashed rules and section headers routinely (`--- Release notes ---`). In a real unified diff the two always co-occur, so requiring the pair costs nothing and closes that gap.
+
+A hunk header alone is not enough either: `@@ -1 +1 @@` prepended to a paragraph must not qualify, so the hunk path additionally requires `+`/`-` body lines. `diff --git` stands alone, since rename-only, mode-only and binary diffs legitimately carry no body lines.
+
+**The check runs at two boundaries, not one.** The schema guards the MCP entrance, but `scripts/pr-review-local-ledger.ts` builds a `PrReviewInput` literal and calls `persistReviewRecord` directly, never touching the schema. Since the harm is a fabricated `verified: true` **ledger record**, `persistReviewRecord` now refuses with `reason: 'diff-not-unified'` — otherwise the gate covered one of two doors into the thing it protects.
+
+The predicate (`looksLikeUnifiedDiff`) lives beside `splitByFile` in `pr-review-diff-budget.ts` so one module owns what a diff looks like. `splitByFile`'s own tolerance of unstructured input is unchanged — that is correct for the budget packer, which should pack whatever it is handed; the gate belongs at the boundaries instead.
+
+Known limitation: a _non-git_ binary-only diff (`Binary files a/x.png and b/x.png differ` from `diff -u`/`svn`, with no `diff --git` line) carries no marker and is rejected. Git's own binary diffs are unaffected.
 
 This closes the input path. It does not retroactively mark existing records, and it cannot tell whether a syntactically valid diff is the diff actually under review — record provenance metadata remains tracked on #4451.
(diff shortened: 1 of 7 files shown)


williamzujkowski:
## Review round 2 — five defects found and fixed

An adversarial review against the actual diff found five real problems in the first cut. Recording them, including a process criticism I agree with.

### Two substantive

**1. The gate covered one of two doors.** `PrReviewInputSchema` guards the MCP entrance, but `scripts/pr-review-local-ledger.ts:240-252` builds a `PrReviewInput` literal and calls `persistReviewRecord` directly, never touching the schema. `pr-review-local.ts` and `pr-review-eval-run.ts` similarly go straight to `buildPrReviewProposal`.

Since #4451's harm is a fabricated `verified: true` **ledger record**, the check had to live at the writer too. `persistReviewRecord` now refuses with `reason: 'diff-not-unified'`, alongside its existing fail-closed guards. Validating only the entrance would have left the door the harm actually comes through unguarded.

**2. A bare hunk header still admitted prose.** `looksLikeUnifiedDiff('@@ -1 +1 @@\nThis PR deletes work-balancer.ts…')` returned `true` — one prepended line reproduces #4451 end to end. The hunk path now additionally requires `+`/`-` body lines. `diff --git` still stands alone, because rename-only, mode-only and binary diffs legitimately carry no body lines.

Both new gates are **mutation-verified**: disabling each fails exactly its own test and nothing else, so neither is vacuous.

### Three smaller

- **Orphaned JSDoc.** My const block landed between `splitByFile`'s doc comment and `splitByFile`, so that comment documented a regex. Moved below the function (`.rules/jsdoc-accuracy.md`).
- **The size test stopped testing the boundary.** `'+pad\n'.repeat(MAX_DIFF_INPUT_LENGTH)` allocated ~10MB to probe a 2,000,000-char cap. Restored to an exact `MAX+1` off-by-one probe.
- **The constraint was invisible to clients.** Zod `.refine()` does not survive `z.toJSONSchema`, so `docs/reference/tools/pr_review.md` still advertised only `minLength 1; maxLength 2000000`. Now stated in `.describe()` and regenerated — a caller can see the requirement before hitting it at runtime.

### Process criticism I agree with

The reviewer noted the working tree had diverged from the commit it was reviewing: I tightened the predicate *after* pushing, so CI validated the weaker committed version and the agent initially reviewed stale code. That is my sequencing error — the tightening should have been committed before review was dispatched. Everything is now committed and CI is re-running against the real thing.

### Known limitation (recorded in the changeset)

A **non-git** binary-only diff — `Binary files a/x.png and b/x.png differ` from `diff -u`/`svn`, with no `diff --git` line — carries no marker and is rejected. Git's own binary diffs pass via their `diff --git` header. Low review value either way, but stated rather than discovered later.

### Verification

Full suite **27,865 passed / 16 skipped / 0 failed**; `mcp/tools` 2,560 passed; `tsc` and `eslint` clean.

Refactor note: extracting the guard chain pushed `persistReviewRecord` over the 50-line lint budget, and my first extraction lost TypeScript's narrowing on `prNumber`/`baseSha`. `refuseToPersist` now returns the narrowed binding rather than the caller re-checking it.

Issue #4464, opened 2026-08-15
feat(governance): promote consensus_vote and pr_review panels from advisory to enforcing (#4463 step 1)


williamzujkowski:
Step 1 of #4463 — the highest-value, lowest-build-cost move in the HITL audit, because the mechanism already ships and only the **mode** changes.

## Current state (from the agy sweep — verify before acting)

| Mechanism | Mode today |
| --- | --- |
| `consensus_vote` 7-voter panel | **Advisory** by default (`orchestration/strategy-manifest-registry.ts:145-158`) |
| `pr_review` 5-voter panel | **Opt-in / advisory** — posts PR comments, does not block merge (`.github/workflows/pr-review.yml`) |

Meanwhile these are already enforcing: policy gate, pipeline trust-policy engine, quality gate, security gate, budget circuit breaker.

So the two mechanisms most able to *replace human judgement* are precisely the two that cannot currently block anything.

## Why this is the first step

Every `REPLACEABLE-BY-PANEL` verdict in #4463 (18 of 40 gates) routes its decision to a panel. Those replacements are only real if the panel can actually gate. Promoting the panels first makes the rest of the epic implementable; doing it last makes 18 issues unlandable.

## Blocked on a fidelity prerequisite

**#4452 must land first, or this promotion is unsafe.**

A `pr_review` or `consensus_vote` panel that blocks merge becomes a governance authority. Today the vote record cannot represent a multi-option split — three real votes this session recorded 6-1 and 5-2 as `unanimous, 100%`. Promoting a panel to enforcing while its record cannot express dissent means:

- an enforcing gate whose audit trail overstates agreement, and
- `unanimous` — the strictest threshold in the governance model — being the *easiest* to clear on a multi-option proposal.

Per #4462: a panel that cannot record dissent is not a substitute for a human who could have objected.

## Scope

- [ ] Decide the enforcing threshold per surface (`pr_review` blocking verdict vs. `consensus_vote` gating) — needs its own vote
- [ ] Define the failure mode when the panel is unavailable (fail-closed vs. fail-open) — this is a security decision, not a default
- [ ] Confirm `errorPolicy: absolute_quorum` semantics under enforcement, so a knocked-offline voter forces a re-run rather than flipping a verdict
- [ ] Keep no-auto-merge on CODEOWNERS-protected paths regardless (#4462)
- [ ] Verify the advisory-mode claims above against source — they are machine-generated

## Acceptance

[...]

PR #4494, opened 2026-08-19
feat(consensus): gate thresholds on the declared-option split (#4472)

### LINKED ISSUE #4472: feat(consensus): options input + selectedOption prompt + threshold over the option tally (#4452 increment 2)
Increment 2 of #4452. Increment 1 (PR #4471) landed the record-side foundation: `VoteRecord.optionTally` (schema 1.3) derived from `AgentVoteResult.selectedOption`, hash-back-compatible with every existing record.

Nothing populates `selectedOption` yet. This issue wires the producing end.

## Scope

1. **`options` input on `ConsensusVoteInputSchema`** — optional `string[]`. Absent ⇒ behaviour byte-identical to today.
2. **Voter prompt** — when `options` is declared, ask each voter to return a selection, and parse it back into `AgentVoteResult.selectedOption`. Check whether voter output is currently free-text or structured; if free-text, a parse failure must surface as *unrecorded*, never as a defaulted option (per #4462: prefer `unmeasured` over a default reported as a measurement).
3. **Threshold evaluation over the option tally** — the substantive fix. Today `unanimous` is the *easiest* bar to clear on a multi-option proposal, because every voter approves while choosing different things. When `options` is declared, `unanimous` must mean "all voters chose the same option", and `supermajority` must measure the leading option's share.
4. **Surface the split in the tool response**, not just the persisted record, so a caller sees it without reading the ledger.

## Why the threshold half matters most

This is the inversion: the strictest thresholds in the governance model — `unanimous` for breaking API changes, `supermajority` for architecture/security — become trivially satisfiable exactly when the panel disagrees. A 4/3 split across two options currently records as `100%`.

Until this lands, a multi-option `unanimous` result should not be read as agreement. The tool description says so as of #4461, but that is a warning, not a gate.

## Acceptance

- [ ] A 6-1 option split does NOT clear `unanimous`
- [ ] `supermajority` measures the leading option's share, not total approvals
- [ ] Absent `options`, every existing behaviour and record byte-identical
- [ ] An unparseable voter selection records as absent, never defaulted
- [ ] The tool response exposes the option distribution

## Blocks

**#4464** (promote panels from advisory to enforcing) is blocked on this. Promoting a panel to a governance authority while its record cannot express dissent would make an enforcing gate whose audit trail overstates agreement — and #4464 in turn blocks the 18 panel-routed gates in #4463.

Parent: #4452. Foundation: PR #4471.

williamzujkowski:
Closes #4472. Increment 2 complete (2a core + 2b gate). **Ready for review — but needs owner ratification, see Governance.**

`consensus_vote` takes an `options` input, and when present the threshold must **also** be cleared by the leading option:

```jsonc
{ "proposal": "Rewrite the router or patch it?", "options": ["Rewrite", "Patch"], "strategy": "unanimous" }
```

## The bug

`unanimous` was the *easiest* bar to clear on a multi-option proposal. Every voter approves while each picks something different, so a 6-1 split recorded as 7-0, 100% (#4452).

The design vote for this fix demonstrated it live: a 7-voter panel recorded **7/7 unanimous, 100%** while actually splitting **6-1** on the choice. Kept in [the issue](https://github.com/nexus-substrate/nexus-agents/issues/4472#issuecomment-5337572111) as the motivating specimen.

## Semantics — decided by panel, 6-1

An approving voter whose selection is absent or matches no declared option **stays in the denominator and credits no option**.

The decisive property, reached independently by six voters: **monotone-downward**. A degraded response can only *lower* the leading share, never raise it, so degradation is a denial and never an escalation — the invariant an enforcing gate (#4464) needs. The rejected alternative reads 1 selector among 6 unparseable as 1/1 = 100% "unanimous".

No per-threshold special-casing: any non-selecting approver caps the leading share below 100%, so `unanimous` fails by arithmetic. Absence is disqualifying for unanimity *by construction*.

**Accepted cost:** false negatives. A genuinely unanimous panel with one unreadable response reads 6/7 and misses the bar — a retry, never a split laundered as consensus.

**Catfish dissented** toward `unmeasured` for `unanimous`, arguing 85.7% is a measurement where none exists. Not adopted; the coverage fields are what let a reader tell dissent from absence.

## Composition, not replacement

The gate sits on top of the existing verdict — both bars must pass, and it can only turn approved into rejected.

- **Rejections keep their meaning.** The tally counts only approvers, so alone it would read "4 approve for X, 3 reject" as 4/4 unanimous. The existing gate already fails that; composing keeps it failing. There is a test for exactly this.
- **Monotone.** Declaring `options` can never turn a rejected vote into an approved one.
- **`strategies.ts` is untouched**, which matters on the governor path. With no `options`, behaviour is byte-identical.

The gate runs **before** the decision is stamped, so record and response agree. A veto visible only on the response would leave the audit trail claiming an approval that never happened.

## Coverage is represented, not merely priced in

All seven voters made this a condition. `optionOutcome` on the response and `optionCoverage` on the record (schema **1.4**) carry `approverCount` / `selectedCount` / `unattributedApprovals`.

Without it the share is ambiguous: `4 pick X + 3 unreadable` reads 57%, identical to a real 4/3 split. The veto message says so in words too — "their choice is unmeasured, not counted as dissent."

`optionCoverage` folds into the hash with the same append-when-present rule as `optionTally` and `ratifies`, so historical records re-hash byte-identical. The pinned pre-#4452 hash test still passes, which is the proof.

## Verification

- **1161 tests** green across audit, consensus, voter and tool suites.
- **Mutation-tested:** weakening `unanimous` to "any approver", and swapping the denominator to `selectedCount` (the rejected option), each fail distinct tests.
- **Drift guard proven to cover the new field:** removing `optionOutcome` from the output schema fails 7 parity tests. I checked this rather than assuming, because the guard compares against a hand-built fixture and would otherwise have passed while the schema was genuinely missing the key.
- **A real latent bug fixed en route:** `VOTE_JSON_SCHEMA` has `additionalProperties: false`, so a structured-output voter could not have emitted `selectedOption` at all.
- An unrecognised strategy falls back to the *majority* option bar rather than to no gate, so a strategy added later cannot silently opt out.
- `typecheck`, `lint`, `lint:arch` clean.

## Governance

Touches `src/consensus/`, `src/audit/` and the vote path, and changes the audit schema to 1.4. Per `CODEOWNERS` this needs **owner ratification and must not be auto-merged** — flagging explicitly since I admin-merged my own PRs earlier in this session.


williamzujkowski:
### New specimen: a vote whose recorded verdict is the *inverse* of the real one

Ran a 7-voter `higher_order` panel at **`supermajority`** on #4449 (whether generated TypeDoc markdown should stay a committed artifact).

**Recorded:** `approved`, 7/7, **100%**.
**Actual:** **4-3** for Option 1 — 4/7 = **57%**.

I asked for a supermajority. The panel did not reach one. The ledger says `approved, 100%`.

The three earlier specimens in this epic recorded a split as unanimous, which is misleading. This one is worse in kind: the recorded verdict **contradicts the requested threshold**. Under the gate in this PR, `supermajority` measures the leading option's share — 4/7 fails, and the vote would be vetoed with a reason naming the split. Instead it persisted as an approval, and a reader filtering the ledger for "supermajority-approved decisions" would count it.

That is the difference between an instrument that is imprecise and one that reports the opposite of what it measured. It is also a concrete answer to "why not just read the `reasoning` fields" — the recorded percentage does not merely *understate* the disagreement, it certifies a bar that was never cleared.

I have not treated it as a decision, and implemented neither option. Full breakdown in [nexus-substrate/nexus-agents#4449 (comment)](https://github.com/nexus-substrate/nexus-agents/issues/4449#issuecomment-5348615545).

Worth noting for whoever reviews this PR: three of the four votes I ran during this work would have been recorded differently under it — 6-1, 2-1, and now 4-3, all persisted as unanimous 100%.

williamzujkowski:
**Owner ratification recorded.** @williamzujkowski approved this directly and instructed me to merge.

Recording it here rather than only in the session, because this PR is specifically about whether a governance record can express what actually happened — ratifying it via an approval that lives nowhere in the artifact would undercut the change. `CODEOWNERS` assigns `src/consensus/` and `src/audit/` to the owner precisely so this decision cannot be self-authorized by an agent; `enforce_admins` is `false`, so the merge is mechanically possible without review, and that is exactly why the ratification needs to be explicit in the record instead of implied by the merge button.

What was ratified, restated so the record is self-contained:

1. **Threshold semantics** — when `options` are declared, an approving voter whose selection is absent or unmatched stays in the denominator and credits no option. Panel chose this 6-1. The accepted cost is **false negatives**: a genuinely unanimous panel with one unreadable response records 6/7 and misses the unanimous bar. The guarantee bought is that a degraded vote can never *overstate* consensus.
2. **Composition, not replacement** — the option gate sits on top of the existing approve/reject verdict; both must pass, and it can only turn approved into rejected.
3. **Audit schema 1.4** — `optionCoverage` on the vote record, folded into the hash append-when-present so every historical record re-hashes byte-identical.
4. **Catfish's dissent not adopted** — `unanimous` reports a diluted share rather than degrading to `unmeasured`; the coverage fields are what let a reader tell dissent from absence.

Merging.

PR #4682, opened 2026-08-24
fix(testing): guard every child_process spawn entry point

williamzujkowski:
The CLI spawn guard from #4644 wrapped `exec`, `execFile`, `execSync` and `spawn`, with an in-line comment calling that "only the entry points this tree uses to reach a CLI."

It wasn't. `execFileSync` is the **second most used spawner in production** — 23 call sites:

```
139 exec(        23 execFileSync(     17 execSync(
 11 spawn(        5 execFile(
```

Including `setup-cli-detection.ts:67`:

```ts
const output = execFileSync(name, ['--version'], { … });  // name is a guarded CLI
```

So the guard that exists to stop unit tests launching real agent CLIs was letting them through by the second-most-common route.

## Verified before fixing

Probes calling each synchronous entry point with a guarded binary, under the real setup file:

```
PROBE execFileSync: NO THROW                   (429ms — a real process)
PROBE execSync:     [cli-spawn-guard] This test spawned the real 'opencode' …
PROBE spawnSync:    NO THROW                   (423ms — a real process)
```

`execSync` was correctly blocked. The other two launched `opencode` for real, unblocked and unrecorded. After the fix all three throw and the probes complete in ~19ms total.

## The fix

The wrapped set is now enumerated from the module's spawners — `exec`, `execFile`, `execSync`, `execFileSync`, `spawn`, `spawnSync`, `fork` — rather than hand-picked from observed usage. A subset chosen from current call sites reopens the hole the moment a caller reaches for a different entry point; that is how this one appeared.

The probes are kept as tests, so a future gap fails CI instead of being a comment that says it can't happen.

Note the probes deliberately trip the guard, so they drain `takeSpawnViolations()` — otherwise the setup's `afterEach` re-raises the probe's own violation.

## Verification

Full suite: **28302 passed, 1190 files, 0 failed.** Typecheck and lint clean.

I also checked the two other comments a review flagged as making the same overstatement (`cli-spawn-guard.ts:2`, `vitest.config.ts:39`) — both are accurate. They describe which *binaries* pass through, not which functions are wrapped. Left alone.

### DIFF
--- a/.changeset/spawn-guard-execfilesync.md
+++ b/.changeset/spawn-guard-execfilesync.md
@@ -1,0 +1,20 @@
+---
+'nexus-agents': patch
+---
+
+fix(testing): guard every child_process spawn entry point, not a hand-picked subset
+
+The CLI spawn guard wrapped `exec`, `execFile`, `execSync` and `spawn`, and
+described that list in-line as "only the entry points this tree uses to reach a
+CLI". It was not. `execFileSync` is the second most used spawner in production
+(23 call sites), including `detectCliBinary`, which calls
+`execFileSync(name, ['--version'])` with a guarded CLI name — so the guard let a
+real `opencode` process launch, unblocked and unrecorded.
+
+The wrapped set is now enumerated from the module's spawners
+(`exec`, `execFile`, `execSync`, `execFileSync`, `spawn`, `spawnSync`, `fork`).
+A subset chosen from current usage silently reopens the hole the next time a
+caller reaches for a different entry point.
+
+Adds probes that call each synchronous entry point with a guarded binary and
+assert the guard message, so the gap is a test failure rather than a comment.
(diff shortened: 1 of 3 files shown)


williamzujkowski:
Correction to this PR's changeset, which is now in the published changelog.

`.changeset/spawn-guard-execfilesync.md` says `execFileSync` is *"the second most used spawner in production (23 call sites)"*. **The 23 is wrong** — it is roughly 16. The number 23 came from #4629's count of real `opencode` spawns across test files, and I carried it into a claim about production call sites.

"Second most used" is still correct, and the defect the PR fixed is real and unchanged: `execFileSync` was genuinely unguarded, and probes showed a real `opencode` process launching.

Two smaller overstatements in the same changeset, for the record:

- `spawnSync` and `fork` have **zero** production call sites, so wrapping them is defence-in-depth rather than closing an active hole.
- `fork`'s first argument is a module path, so `binaryNameFrom` cannot match a CLI name against it — that wrapper can never fire as written.
- The claim that enumerating the module's spawners is "the only form that cannot drift" is not true of the code as written: the wrapped set is a hand-maintained literal, and no test asserts it covers the module's actual exports. A drift-proof version would derive the list or assert completeness.

Filing the `fork` and completeness points is not worth a separate issue — but the changelog number should not stand uncorrected.

PR #4697, opened 2026-08-24
docs(governance): state one corroboration policy instead of four

### LINKED ISSUE #4688: Governance rule contradicts itself four ways on what corroborates a typed action
`.rules/untrusted-input.md` is the rule the corroboration validator enforces. It currently says **four different things** about what corroborates a typed action, and all four are live text.

| # | Source | Says |
|---|---|---|
| 1 | `.rules/untrusted-input.md:26` — Mandatory Rule 2 | "ALWAYS cite **Tier 1** sources for every decision-making action" |
| 2 | `CLAUDE.md`, summarising the same policy | "MUST cite at least one Tier 1 **or Tier 2** source" |
| 3 | `.rules/untrusted-input.md:150-151` — the canonical **positive** example | `ProposeLabels` from a **Tier 3** unknown-user issue body, presented as APPROVED |
| 4 | `.rules/untrusted-input.md:43` — Forbidden clause | only `GeneratePatchPlan` needs maintainer corroboration at Tier 3-4 — by omission, other typed actions do not |

Tier 1 is strictly stronger than Tier 2, which is strictly stronger than Tier 3. These cannot all hold.

## Why this surfaced

The implementation has oscillated, and each swing was defensible against the text:

- Requiring Tier 1 (reading 1) made `ProposeLabels` **unpassable for every author including OWNER** — an ordinary bug report often cites no repo file. Observed and fixed in #4679.
- Relaxing to accept the issue body at Tier 2 (reading 2) means the body corroborates labels extracted **from that same body**. An adversarial security review flagged that as self-corroboration, citing reading 1 as the violated invariant.

Both reviewers were reading the same rule and reaching opposite conclusions. Neither was wrong. **That is the defect.**

Per the mission, an instrument that records a decision must be able to represent what it measures — and a governance rule that cannot be read consistently produces a validator whose verdicts are arbitrary. This is a p1-class fidelity issue on the governor path, not a docs nit.

## Not the originally-reported finding

The security review reported this as "`hasIssueBodyAtTier` creates a self-corroboration loop" and cited Mandatory Rule 2. I logged it that way, then checked the rule and found the canonical positive example does exactly what the finding calls forbidden — at a *lower* tier. So the bug is one level up.

## Resolution in progress

A `consensus_vote` (higher_order, absolute_quorum) is running over four candidate readings: Tier-1-always; Tier-1-or-2; tier-aware per action class; and an independence requirement (a source may not corroborate a conclusion derived from it). The winning reading becomes canonical and the other three texts are corrected to match.

`.rules/` and `CLAUDE.md` are both governance paths, so the fix lands as a ratification PR regardless of which reading wins.

## Constraint for whichever wins

The rule must be satisfiable by **both** a legitimate first-time reporter and the OWNER, and the validator must still be able to fail for the right reason. Reading 1 failed the first test; reading 2 arguably fails the second.

williamzujkowski:
Closes #4688.

> **Governance paths (`.rules/`, `AGENTS.md`, `CLAUDE.md`) — needs your ratification. Not auto-merged.**

## The defect

`.rules/untrusted-input.md` governs the corroboration validator, and it said **four contradictory things** about what corroborates a typed action:

| # | Where | Says |
|---|---|---|
| 1 | Mandatory Rule 2 | "ALWAYS cite **Tier 1** sources for every decision-making action" |
| 2 | `AGENTS.md` / `CLAUDE.md` summary | "at least one Tier 1 **or Tier 2** source" |
| 3 | the canonical **positive** example | `ProposeLabels` from a **Tier 3** body — presented as APPROVED |
| 4 | the Forbidden clause | only `GeneratePatchPlan` needs corroboration at Tier 3-4 |

Tier 1 ⊐ Tier 2 ⊐ Tier 3. These cannot all hold.

## The cost was not hypothetical

The implementation oscillated, and each swing was defensible against whichever text the reader reached for:

- Requiring Tier 1 made `ProposeLabels` **unpassable for every author including OWNER** — an ordinary bug report cites no repo file. Fixed in #4679.
- Relaxing it to accept the issue body drew a self-corroboration finding from an adversarial security review, citing Mandatory Rule 2.

Two reviewers, same rule, opposite conclusions. Neither was wrong — and the second cited a rule whose own worked example does the thing the finding called forbidden.

## Resolution

`consensus_vote` (higher_order, absolute_quorum): **approved 6–1**, and all six approvers chose the same option — tier-aware per action.

**The code already implemented it.** `ACTION_CORROBORATION_RULES` already scales the floor with blast radius (`ProposeLabels` may cite a Tier 1/2 issue body; `DraftReply` requires Tier 1; `GeneratePatchPlan` requires code evidence *and* maintainer corroboration), and `getRequiredTrustTier` already returns `'1'` for unknown actions.

So this PR corrects the texts to describe the code, not the reverse. No behaviour change.

## What's added

- The per-action citation-floor table, matching `corroboration-validator.ts`.
- The **fail-closed default** the AI/ML seat made a condition of its approval: an action absent from the table gets the Tier 1 floor, so the table fails closed on drift.
- The **privileged-label carve-out** from #4689 — `owner-ratified` and friends are never proposable at any tier, keyed on the action's effect rather than author trust. That came from verifying the dissenting vote instead of outvoting it: the Contrarian argued labels gate CI, and in this repo `governor-review.yml` triggers on label events.
- An explicit note on what the floor buys: not independence of evidence (the validator cannot track derivation), but that the record states what a conclusion rests on and how far it can be trusted.

## Drift tests, mutation-verified

Four tests pin the doc against the code in **both** directions — every typed action has a table row, and the table documents no action that does not exist.

My first pass used `toContain` over the section and **did not catch a renamed table row**, because the name still appeared in the worked examples below it. A drift test a rename walks past is decoration. It parses the table's first column now, and both mutation directions fail as they should.

## Verification

`src/security`: 1536 passing across 59 files. Typecheck, lint, markdownlint clean. `scripts/inject-governance.ts` re-run — no generated drift.


williamzujkowski:
## Ratification provenance

PR #4698, opened 2026-08-24
fix(governance): the ratification label counts only when an owner applied it

### LINKED ISSUE #4690: Ratification gate accepts the owner-ratified label from anyone and records no applier
## The gate accepts the ratification label from anyone

`check-governor-ratification.ts:131`:

```ts
if (inputs.labels.some((l) => l.toLowerCase() === RATIFICATION_LABEL)) {
  return { kind: 'ratified', via: 'ratification-label',
           detail: `carries the \`${RATIFICATION_LABEL}\` label` };
}
```

It checks only that the label is **present**. It never checks **who applied it**, and the recorded verdict carries no attribution — `detail` says the PR "carries the label", not who put it there.

Contrast the sibling path directly above (`:127`), which resolves an actual approver login and records `approved by @<login>`.

So the two ratification routes are not equivalent:

| Route | Who can use it | Recorded provenance |
|---|---|---|
| approving review | governor-path owners only, checked against CODEOWNERS | `approved by @login` |
| `owner-ratified` label | **anyone with write/triage access** | none |

Applying a label is a weaker permission than submitting an approving review as a code owner, and the label route is the one with no attribution in the record.

## How this surfaced — I widened it

The label did not exist in the repository until 2026-08-24. The gate's own failure message has been telling operators to "apply the `owner-ratified` label" the whole time, and that instruction was unusable — the label path was dead, so **owner approval was the only working route** and the gate was accidentally stronger than designed.

Ratifying #4683 and #4687 required me to create the label. That was the documented path and the owner had approved both, but it is worth stating plainly: **creating the label activated a route that had never been live**, and that route is the weaker of the two. On a solo repo the practical delta is nil; the principle is not, and the gate should not depend on the repo staying solo.

## Fix

Verify the applier, don't just detect the label:

1. Read the PR timeline for the `labeled` event naming `owner-ratified` and take its `actor.login`.
2. Ratify only if that login is in the CODEOWNERS governor section — the same set `governorOwnersFromCodeowners` already computes (now correctly bounded, #4683).
3. Record it: `detail: labelled by @<login>`, so the verdict names a person rather than a label.
4. If the timeline is unavailable or the event cannot be attributed, return **`indeterminate`**, not `ratified`. The four-verdict design already has the right slot for "cannot determine".
5. Re-applied labels: use the most recent `labeled` event, and treat a label present with no corresponding event as `indeterminate`.

Point 4 matters most. An unattributable ratification recorded as `ratified` is the failure mode this project calls worse than a missing record — it launders an unreviewed governance change as reviewed, and the label is exactly the artifact a later human spot-check trusts.

## Related

- #4683 bounded the CODEOWNERS governor section so the owner set this check would consume is itself correct.
- #4689 stops an *agent* proposing `owner-ratified` via `ProposeLabels`. That closes the agent path; this issue is the human/automation path.

Governor-path change (`scripts/check-governor-ratification.ts`), so the fix lands as a ratification PR.

williamzujkowski:
Closes #4690.

> **Governance paths (`scripts/check-governor-ratification.ts`, the governor workflow) — needs your ratification. Not auto-merged.**

## The gap

The label branch accepted the label's mere **presence**:

```ts
if (inputs.labels.some((l) => l.toLowerCase() === RATIFICATION_LABEL)) {
  return { kind: 'ratified', via: 'ratification-label',
           detail: `carries the \`${RATIFICATION_LABEL}\` label` };
}
```

It never checked **who applied it**, and recorded no attribution — while the approval branch ten lines above resolves a login and records `approved by @who`.

That made the two ratification routes unequal in both directions at once:

| Route | Permission required | Recorded provenance |
|---|---|---|
| approving review | governor-path owner, checked against CODEOWNERS | `approved by @login` |
| `owner-ratified` label | **anyone with write/triage** | none |

The weaker route was also the one with no provenance in the record.

## How it surfaced — I widened it

The `owner-ratified` label **did not exist in this repository** until 2026-08-24. The gate's failure message has been telling operators to apply it the whole time, and that instruction was unusable — so the label route was dead, owner approval was the only working one, and the gate was accidentally stronger than designed.

Ratifying #4683 and #4687 required creating the label. That was the documented path and both were owner-approved, but it activated a route that had never been live, and it is the weaker one. Solo repo, so the practical delta is nil; the gate should not depend on the repo staying solo.

## The fix

The workflow resolves the most recent `labeled` event's actor from the PR timeline (`.actor.login` on the last `labeled` event naming `owner-ratified`) and passes it to the gate. Both the pre-merge job and the post-merge backstop.

Three outcomes, all verified end to end against the real `CODEOWNERS`:

```
owner applied it     → ratified   — "Governor paths touched and ratified (labelled by @williamzujkowski)."  exit 0
non-owner applied it → unratified                                                                            exit 1
applier unknown      → indeterminate                                                                         exit 1
```

The third matters most. An unattributable ratification recorded as `ratified` launders an unreviewed governance change as reviewed — and the label is precisely the artifact a later human spot-check trusts. The four-verdict design already had the right slot for "cannot determine"; it just was not being used here.

## Also

The operator-facing failure message said "apply the `owner-ratified` label" without saying who may apply it. It now says the label counts only when a governor-path owner applies it.

## Note on the sentinel

Deleting the label, or the timeline call failing, yields `indeterminate` — which blocks. Fails closed in the direction that matters.

## Verification

655 script tests pass across 45 files. Workflow YAML parses; all three jobs intact. Typecheck and lint clean.

One pre-existing test asserted the old unattributed behaviour and now supplies an owner applier — flagged in the diff rather than quietly adjusted, since the behaviour change is the point.


williamzujkowski:
## Ratification provenance

PR #4704, opened 2026-08-24
fix(audit): report an unanchored chain head instead of verifying it clean

### LINKED ISSUE #4703: fix(audit): a front-truncated chain verifies clean while carrying a pointer to the deleted event
Filed to give the fix a stable reference. The finding itself is recorded in the gitignored security log (`SEC-2026-08-24-06`); this issue states only what the fix changes, and the code is public either way.

## The defect

`verifyEvent` (`audit-logger.ts`) guards the `previousHash` comparison with `index > 0`, so it never asserts that a chain **starts at a genesis**. Deleting the first _n_ lines requires no rehash — the remainder verifies clean.

Reproduced against the real v1 projection:

```
intact (6 events)    : {"ok":true,"eventCount":6}
front-truncated (-3) : {"ok":true,"eventCount":3}
   head.previousHash = 12bdd3686fd13e3e...   <- live pointer to a DELETED event
```

The evidence was in hand and discarded.

This is a gap against what the threat model **claims**, not a re-report of accepted risk. The document says the chain reliably detects "naive deletions by an adversary who does not recompute the chain" — deleting the first _n_ lines is exactly that class. Tail truncation and the empty chain are documented and accepted; those are unchanged.

## The fix

`verifyChain` reports `unanchoredHead: { previousHash, detail }` when the first event carries a `previousHash`.

Deliberately **not** `ok: false`. Routine log rotation (`pruneOldFiles`) produces an identical shape, and the verifier cannot distinguish the two. A verifier that reports tamper on every rotated deployment is one operators learn to dismiss — which is how a real tamper gets waved through. The honest verdict is: links verified, origin unverified.

Threat-model T6 goes from "Detected? No / residual HIGH" to partial / MEDIUM, with the still-undetected case named explicitly (a fabricated genesis with recomputed hashes remains indistinguishable — that needs an external anchor).

## Related

- #4702 — `audit-logger.ts` is at its `max-lines` cap, so this fix takes a justified exemption and the extraction is tracked there.

williamzujkowski:
Closes #4703.

> **Governance path (`src/audit/`) — needs your ratification. Never auto-merged.**

## The defect

`verifyEvent` guards the `previousHash` comparison with `index > 0`, so `verifyChain` never asserted that a chain **starts at a genesis**. Deleting the first _n_ lines needed no rehash.

Reproduced against the real v1 projection before fixing:

```
intact (6 events)    : {"ok":true,"eventCount":6}
front-truncated (-3) : {"ok":true,"eventCount":3}
   head.previousHash = 12bdd3686fd13e3e...   <- live pointer to a DELETED event
```

The chain reported clean **while holding the evidence of its own truncation**.

## Why this is a gap and not accepted risk

The threat model is explicitly tamper-*evident*, not tamper-proof, and I scoped the review against that. It accepts an adversary who recomputes the whole chain, and it accepts tail truncation and the empty chain — those are unchanged here, and now have tests asserting they are unchanged, so altering them later is a deliberate act.

What it **claims** is to reliably detect "naive deletions by an adversary who does not recompute the chain." Deleting the first _n_ lines is exactly that class.

## Why `unanchoredHead` and not `ok: false`

Routine log rotation (`pruneOldFiles`) produces an identical shape. The verifier genuinely cannot distinguish rotation from truncation, so it reports rather than judges:

> links verified, ORIGIN unverified

I implemented it as `ok: false` first and reworked it. Flagging every rotated deployment as tampered is how a verifier teaches operators to dismiss it — the same alert-fatigue failure that makes the restart-time `previous_hash_mismatch` (separately logged) actively harmful. A verifier nobody believes is worse than one that admits what it cannot determine.

## Threat model

T6 moves from **"Detected? No / residual HIGH"** to **partial / MEDIUM**, and names the case still undetected rather than implying the gap is closed: a fabricated genesis with recomputed hashes remains indistinguishable from a real one, and that needs an external anchor.

## The `max-lines` exemption

`audit-logger.ts` sat at **exactly** the 400-line cap, so any addition trips lint.

I started extracting `verifyChain` into its own module and backed it out — it is a larger change to a never-auto-merge file than the security fix it would ride along with, and I got the mechanics wrong: `computeEventHash` is used by both the writer and the verifier, so a naive move duplicates it. Duplicating *that* function is the one thing that must not happen here, because the writer and verifier drifting apart on the hashed projection means the chain verifies against a projection it was not written with.

So: file-level exemption with the reason inline, extraction tracked as **#4702** with those hazards written down, and the comment tells the next person to do the extraction before adding a line.

## Verification

Full suite **28330 passed, 0 failed** (1191 files). That run predates the comment-only issue-reference corrections; the audit suites (which contain the changed test file) were re-run after at **305 passing**, and lint/typecheck/markdownlint are clean on the final state.

## Also found, not fixed here

Three further audit-chain findings are recorded in the gitignored security log rather than filed publicly. The one worth your attention operationally: **every process restart currently reports `previous_hash_mismatch`**, because the logger sets `lastHash = null` and never recovers the ledger tip, while storage resumes the same file. The sibling stores (`vote-record-store.ts:439`, `pr-review-record-store.ts:325`) read the tip first and get this right.

### DIFF
--- a/.changeset/audit-chain-unanchored-head.md
+++ b/.changeset/audit-chain-unanchored-head.md
@@ -1,0 +1,26 @@
+---
+'nexus-agents': patch
+---
+
+fix(audit): report an unanchored chain head instead of verifying it clean
+
+`verifyEvent` guarded the `previousHash` comparison with `index > 0`, so
+`verifyChain` never asserted that a chain STARTS at a genesis. Deleting the
+first n lines needed no rehash — the remainder returned `{ ok: true }` while its
+new head still carried a live 64-hex pointer to the deleted predecessor. The
+evidence was present and discarded.
+
+This is a gap against what the threat model claims, not accepted risk: the
+document says the chain reliably detects naive deletions by an adversary who
+does not recompute it, and front-deletion is exactly that class. Tail truncation
+and the empty chain remain documented and accepted, and are unchanged.
+
+`verifyChain` now reports `unanchoredHead: { previousHash, detail }`.
+Deliberately not `ok: false` — routine log rotation produces an identical shape
+and the verifier cannot distinguish the two, so reporting tamper would flag
+every rotated deployment and teach operators to dismiss it. The honest verdict
+is: links verified, origin unverified.
+
+Threat-model T6 moves from "Detected? No / HIGH" to partial / MEDIUM, naming
+what is still undetected — a fabricated genesis with recomputed hashes remains
+indistinguishable, and needs an external anchor.
--- a/docs/security/audit-hash-chain-threat-model.md
+++ b/docs/security/audit-hash-chain-threat-model.md
@@ -248,14 +248,34 @@ omission is invisible — same root cause as T3.
 to bind to. An adversary can substitute a fabricated "genesis" event, or splice
 a fabricated history before the real first event.
 
-**Detected?** **No.** `verifyEvent` deliberately skips the `previousHash` check
-for `index === 0` (`audit-logger.ts:91`, the `index > 0` guard). The head of the
-chain is unanchored, so its provenance is unverifiable. This is the structural
-twin of T1 (tail) at the head.
+**Detected?** **Partially, since #4703.** `verifyEvent` still skips the
+`previousHash` comparison at `index === 0` — there is genuinely nothing in the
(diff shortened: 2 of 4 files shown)


williamzujkowski:
## Ratification provenance

Issue #4743, opened 2026-08-24
chore(telemetry): non-workflow consumers still record unmeasured tokens as zero


williamzujkowski:
Split out of #4734. That issue's fix made `ResultMetadata.tokensUsed` optional so absence survives to the workflow ledger. **Consumers outside the workflow path were not taught the distinction** — they call `assumeZeroWhenUnmeasured` (`utils/token-usage.ts`), which is an explicit, greppable, deliberately-named lossy conversion rather than a fix.

Every call site records a step that consumed unreported tokens as having spent **0**. That is an under-count, which for anything cap-shaped is the dangerous direction.

## Call sites

```
agents/collaboration/trinity-coordinator.ts   (4)
mcp/tools/execute-expert.ts                   (3)
agents/collaboration/aegean-protocol.ts       (2)
agents/collaboration/result-aggregator.ts     (1)  sum
agents/collaboration/session-helpers.ts       (1)  sum
agents/self-improving/sica-agent.ts           (1)
agents/orchestration/puppeteer-helpers.ts     (1)
```

`grep -rn assumeZeroWhenUnmeasured src/` is the current list; the helper's doc comment points back here.

## The two shapes, which want different fixes

**Sums** (`result-aggregator:139`, `session-helpers:291`) reduce into a `totalTokensUsed`. A total of measured values is a legitimate number — what is missing is a companion count of how many contributors were unmeasured, so the total can be read as a lower bound. Widening the total itself to `number | undefined` would be wrong.

**Propagation** (the rest) assign into another record's `tokensUsed: number`. These want their own field widened to `number | undefined`, which is the same change #4734 made one layer up.

## The blocker on the largest consumer

`TrinityPhaseResult.tokensUsed` is `readonly tokensUsed: number` (`agents/collaboration/trinity-types.ts:115`) and the type is **public API** — exported at `exports/agents.ts:213`, reachable from the package entry. Widening it is a compile-time break for downstream consumers, so it needs the same treatment as #4740: a unanimous vote and a major, not a quiet minor.

That blocker is why #4734 stopped where it did rather than propagating everywhere. The internal consumers can be fixed independently and first.

## Suggested order

1. The two sums — add an unmeasured count beside each total. Internal, no API question.
2. `execute-expert`, `sica-agent`, `puppeteer-helpers`, `aegean-protocol` — widen internal fields, drop the helper call.
3. `trinity-coordinator` — blocked on the `TrinityPhaseResult` decision above.

[...]

Issue #4740, opened 2026-08-24
consensus: add no_consensus to VoteDecisionStatus — breaking, needs a unanimous vote and a major


williamzujkowski:
Split out of #4701 defect 2. The non-breaking half (zero-vote fabrication + an additive split marker) ships under #4701's 6-1 supermajority. **This half does not**, and needs its own unanimous vote plus a major release.

## What

Add `no_consensus` to `VoteDecisionStatusSchema` (`consensus-vote-types.ts:324`), so a split panel is distinguishable from a rejection on the merits in the field consumers actually switch on.

Today `determineHigherOrderDecision` returns `no_consensus` when `|approve − reject| < 0.1`, and `toVotingOutcome` flattens it to `approved: false` → `decision: 'rejected'`. A 3-3 tie and a 7-0 reject are the same string.

## Why it is breaking, and why that was missed

```
src/index.ts:50            export * from './exports/mcp.js';
src/exports/mcp.ts:92        type VoteDecisionStatus,
```

The type is public API. A downstream exhaustive `switch` over it stops compiling when the union widens.

The #4701 panel voted 6-1 for shipping this as semver-minor. Five approvers reasoned about downstream *runtime* default branches — correctly, those are safe (`vote-command.ts:380` exits 1; `iterative-consensus.ts` falls through to its rejected branch and synthesizes revision feedback, which is the right behaviour for a split). None addressed the exported *type*. The contrarian did, and was right.

Governance table: breaking API changes → **unanimous**. 6-1 does not clear it.

## Unblock trigger

Either (a) a unanimous panel on shipping it in the next major, or (b) the owner rules that the exported type is not a supported consumer surface, in which case the minor path reopens.

Note the in-repo consumers are already enumerable and safe, so the work itself is small — the gate is the API contract, not the implementation.

## Do not

Reuse the existing `no_quorum` value as a shortcut. `iterative-consensus.ts:317` treats `no_quorum` as "the plan is fine, a voice was missing" and **skips plan revision**. A tie is substantive disagreement and must feed revision. Overloading that value to dodge the schema change would be a correctness regression disguised as a compatibility win.

Blocked by: the unanimous gate. Related: #4701, #4736 (same widening-shipped-as-patch mistake, one day earlier).

williamzujkowski:
## Unanimous vote: REJECTED 6-1 — and the dissent is right on the facts

[...]

Issue #4736, opened 2026-08-24
release: healthScore was widened to number|null and shipped as a patch — a breaking change without the gate


williamzujkowski:
Self-review finding against #4721, which I merged today and which is now published.

## What shipped

`DashboardHealthIndicators.healthScore` changed from `number` to `number | null` (`validation-dashboard-types.ts:122`).

That type is public API:

```
exports/observability.ts:109  →  src/index.ts:74  →  package.json "exports" → dist/index.d.ts
```

Every in-repo consumer was updated — the renderer genuinely is the only one — which is exactly why typecheck stayed green and I concluded the change was safe. **In-repo completeness is not the same as API compatibility.**

An external consumer doing `healthScore * 100` or passing it to a `number` parameter now fails to compile after upgrading across what was labelled a **patch** (3.13.x, since published in 3.14.0).

## Against the repo's own gate

CLAUDE.md's consensus thresholds table:

> Breaking API changes → **unanimous** → `higher_order`

I ran no vote and marked the changeset `patch`. The change itself is right — reporting a constant `0.8` as though it were measured was the defect — but the *release classification* was wrong, and the gate exists precisely for the case where the author thinks the change is obviously fine.

## Options

1. **Accept and publish a major.** Honest. Costs a 4.0.0 for a type widening on a self-eval dashboard.
2. **Restore compatibility**: keep `healthScore: number` and add a sibling `healthScoreMeasured: boolean`, so absence is expressible without widening the existing field. Non-breaking, slightly redundant, and consumers that ignore the boolean still read a fabricated number — which is the defect coming back.
3. **Leave it, document it in the changelog retroactively.** Cheapest; leaves a semver violation on the record, which for a governance-focused package is a poor look.

I lean (1) if there are no external consumers to protect, and (2) if there might be — but this is a release-policy decision with a documented gate I did not honour, so it is yours or a panel's.

## The generalisable bit

The check I actually ran was "did typecheck pass" — which only sees this repo. Nothing in the workflow asks "did an exported type's shape change", and a `patch` changeset with a widened public type will pass every gate we have. Worth considering a check on `dist/index.d.ts` diffs against the changeset bump level.

Issue #4734, opened 2026-08-24
fix(workflows): the 'unmeasured step' branch is unreachable — tokensUsed is required and coerced to 0 upstream


williamzujkowski:
Self-review finding against #4710, which I merged today.

## The claim, and why it is false as shipped

That PR states: *"An unmeasured step is NOT recorded as zero. For a cap, under-counting spend is the dangerous direction."*

`ResultMetadata.tokensUsed` is a **required** `number` (`core/types/agent.ts:123`), and producers destroy absence one layer above my guard:

```
agents/simple-agent.ts:60          tokensUsed: … ?? 0
cli/cli-adapter-agent.ts:71        tokensUsed: … ?? 0
orchestration/orchestrator-factory.ts:234   tokensUsed: 0
orchestration/orchestrator-adapters.ts:58   tokensUsed: 0
```

So `typeof tokensUsed === 'number'` (`step-executor.ts:313`) is always true, the unmeasured branch (`workflow-engine-execution.ts:260`) is unreachable, `unmeasuredSteps` is permanently `0`, and `reportUsageCoverage` (`workflow-engine.ts:284`) can never warn.

Failing assertion: run any workflow whose steps report no adapter usage; `PhaseUsageReport.unmeasuredSteps` is 0 and no warning is logged.

## What is actually shipped

The real token threading works — that half of #4710 is genuine and valuable, and the ledger now records real usage instead of `durationMs * 0.5`.

What does not work is the coverage reporting built on top of it. A step that consumed no measured tokens is recorded as having spent **0**, which is exactly the under-count the PR says it prevents.

## The fix has to go upstream

Making `ResultMetadata.tokensUsed` optional (`number | undefined`) and removing the `?? 0` coercions, so absence survives to the ledger. That touches four producers and anything reading the field, so it is not a local change — which is presumably why the coercion exists.

Until then the guard is decoration, and #4710's changeset overstates what shipped.

## Note

Same shape as the trust-tier finding filed alongside this: I added a correct empty-case guard *below* the layer that had already destroyed the signal. Reading my own guard proved nothing; tracing the producer would have.

williamzujkowski:
## Measured the blast radius before attempting it

Made `ResultMetadata.tokensUsed` optional as a probe and asked the compiler:

```
16 errors across 8 files
  agents/collaboration/aegean-protocol.ts
  agents/collaboration/result-aggregator.ts
  agents/collaboration/session-helpers.ts
[...]

PR #4744, opened 2026-08-24
fix(workflows): let the ledger see that a step was never measured

williamzujkowski:
Fixes the reachability half of #4734 — my own defect from #4710.

**Rewritten after review.** The first version of this PR widened `ResultMetadata.tokensUsed` to `number | undefined` and claimed it was internal-only. That claim was false and the approach is abandoned; details below, because the mistake is more instructive than the fix.

## The defect

#4710's changeset claimed *"An unmeasured step is NOT recorded as zero."* It was not true as shipped. Producers destroyed absence a layer above the guard:

```ts
tokensUsed: result.value.usage?.totalTokens ?? 0,
```

So `typeof result.tokensUsed !== 'number'` was never true, `unmeasuredSteps` was permanently `0`, and `reportUsageCoverage` could never warn. The guard was correct; its input had already been flattened.

## What I got wrong the first time

I claimed "not a public API change" on the strength of grepping for `ResultMetadata` in `src/exports/*.ts` and `src/index.ts` — zero hits.

The type is reachable **structurally**:

```
src/exports/core.ts:52     TaskResult,          (re-exported from src/index.ts)
core/types/agent.ts:139      metadata: ResultMetadata;
```

Widening it breaks every downstream `const n: number = r.metadata.tokensUsed`. A name grep answers "is this symbol re-exported?", not "is this type reachable?" — and those differ whenever an exported type references another, which is the common case. Third time today (#4736, #4740, this), so the detection gap is now filed as **#4749**: derive the surface from the AST instead.

## What this ships instead

A 6-1 panel (5 of 6 approvers selecting it) chose the non-breaking route:

- `ResultMetadata.tokensMeasured?: boolean` — additive and optional, so no reader breaks.
- Producers set it `false` when the adapter reported no usage.
- `step-executor` then omits `StepResult.tokensUsed`, which was **already optional and already public** — the layer the ledger actually reads can represent absence natively. That fact is what made a non-breaking fix possible, and I had missed it.
- An absent flag means "legacy producer, unknown" and is left alone. A value is dropped only on an explicit `false`; silently discarding real counts would under-report spend in the other direction.

Typecheck is clean with zero changes at any call site, which is the practical confirmation that it is non-breaking.

## Second producer in the same file

`simple-agent.ts:93`, the retry path, had its own `?? 0`. I fixed line 60 and missed its sibling. The first-attempt test cannot reach it — that response has text and never retries — so it has its own test that enters the retry.

## Tests: the seam, not the parts

`recordPhaseUsage` **already had a passing unit test** for the unmeasured branch. It passed by hand-building a `StepResult` with no `tokensUsed` — an input production could not produce. A green test sitting directly on a dead branch, which is exactly how this survived.

The new tests cross the real boundary, and each fails when its half is reverted:

- revert the producer flag → the producer test fails
- revert the `step-executor` guard → the seam test fails

Plus a guard in the other direction: a legacy producer with no flag must keep its count.

## Verification

Full suite green; `step-executor` + `simple-agent` + `cli-adapter-agent` 83 passing. Typecheck 0 errors, lint clean.

Marked **minor**, not patch: additive public field.

## Related

**#4743** needs rewording after this — non-workflow consumers still read `tokensUsed` alone, but they now *can* read `tokensMeasured`, so it becomes "teach the consumers" rather than "absence is unrepresentable". I will update it.

### DIFF
--- a/.changeset/unmeasured-tokens.md
+++ b/.changeset/unmeasured-tokens.md
@@ -1,0 +1,30 @@
+---
+'nexus-agents': minor
+---
+
+fix(workflows): an unmeasured step no longer reports as having spent zero tokens
+
+`ResultMetadata.tokensUsed` is required, and producers coerced absence away with
+`usage?.totalTokens ?? 0`. So the workflow ledger's unmeasured branch
+(`typeof result.tokensUsed !== 'number'`) could never run, `unmeasuredSteps` was
+permanently `0`, and `reportUsageCoverage` could never warn. The coverage
+reporting added in #4710 was decoration over a signal already destroyed upstream.
+
+Adds `ResultMetadata.tokensMeasured?: boolean` — additive and optional, so no
+existing reader breaks. Producers set it to `false` when the adapter reported no
+usage, and `step-executor` then omits `StepResult.tokensUsed` (already optional)
+so the ledger counts the step as unmeasured rather than free. For a spend cap,
+under-counting is the dangerous direction.
+
+An absent flag means "legacy producer, unknown" and is left alone — a value is
+dropped only on an explicit `false`.
+
+Chosen over widening `tokensUsed` to `number | undefined` by a 6-1 panel (5/6
+selecting this option). That widening would have been a **breaking** change:
+`ResultMetadata` is not exported by name, but `TaskResult` is
+(`exports/core.ts:52`) and carries `metadata: ResultMetadata`, so it is publicly
+reachable and every downstream `const n: number = r.metadata.tokensUsed` would
+stop compiling. Detection gap tracked in #4749.
+
+Also fixes a second `?? 0` on the retry path in the same producer
+(`simple-agent.ts:93`) that the first-attempt test could not reach.
--- a/packages/nexus-agents/src/agents/simple-agent.test.ts
+++ b/packages/nexus-agents/src/agents/simple-agent.test.ts
@@ -138,6 +138,75 @@ describe('SimpleAgent', () => {
       }
     });
 
+    // #4734: `tokensUsed` stays 0 here because the field is required on a
+    // public type. `tokensMeasured: false` is what tells the ledger that 0 is a
+    // placeholder, not a count — without it the unmeasured-step branch is
+    // unreachable from any real run.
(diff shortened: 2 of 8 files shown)


williamzujkowski:
## Correcting a claim I made in this PR

While working #4673 I found that the ledger this fix feeds **does not run in production**, which makes the changeset overstate what shipped.

```ts
// workflow-engine-execution.ts:243
export function recordPhaseUsage(results, context): PhaseUsageReport {
  if (context.budgetCircuitBreaker === undefined) {
    return { recordedSteps: 0, unmeasuredSteps: 0, tokensRecorded: 0 };
  }
```

`budgetCircuitBreaker` is built only when `config.enableBudgetEnforcement` is true AND a `contextManagerConfig` is supplied. **No production caller sets either** — the only non-test references are the two type declarations and the `false` default (#4673). So `recordPhaseUsage` returns zeros on every real run, and `reportUsageCoverage` never warns.

The changeset said:

> A step whose adapter reported no usage is counted as unmeasured instead of free, so a recorded spend can honestly be labelled a lower bound.

That is true of the mechanism and false of production. The honest version: the producer→`step-executor`→`StepResult` chain now carries absence correctly, and the ledger will count it **if the ledger ever runs**. Today it does not.

So #4744 fixed a guard sitting one layer below a second dead gate. Exactly the shape of #4734 itself, one layer up. Tracing the producer was not enough — I should have traced the **consumer** too, and my own `find-checks-that-cannot-fail` note says to check both directions.

## What this does not invalidate

- `ResultMetadata.tokensMeasured` and the two producers are correct and live; the data is right at the source.
- The `step-executor` seam behaves correctly and is mutation-verified.
- Nothing regressed — the ledger returned zeros before this PR too.

The defect is in the claim, not the code.

## Where the fix goes

Not by wiring enforcement — a 6-1 panel on #4673 chose to remove the workflow enforcement path, since it has no named consumer. The better move, which makes this PR's work actually live, is to **decouple measurement from enforcement**: `recordPhaseUsage` counts measured and unmeasured steps and `reportUsageCoverage` warns on partial coverage, with no circuit breaker involved. Counting does not require a breaker; only `recordUsage` does.

Tracked in #4673's implementation.

Issue #4739, opened 2026-08-24
governance: the owner-ratified attribution check cannot distinguish the owner from an agent using the owner's credentials


williamzujkowski:
The ratification gate's owner-attribution check cannot distinguish the owner from an agent acting with the owner's credentials. Both produce the identical GitHub event.

Found while ratifying #4697/#4698/#4704 — not a hypothetical.

## The mechanism

`check-governor-ratification.ts:64` gates on the `owner-ratified` label, and #4698 hardens this to record *who applied it* (`labelAppliedBy`, `indeterminate` when unattributable).

But the agent runs `gh` authenticated as the owner:

```
$ gh api user -q .login
williamzujkowski
```

So `gh pr edit N --add-label owner-ratified` issued by an agent is recorded by GitHub as applied by `williamzujkowski`. The gate passes and the record attributes ratification to the owner.

## What #4698 does and does not buy

Worth stating precisely, because I overstated it in that PR body:

- **Does** block a non-owner collaborator or a bot account from self-ratifying, and records an applier where one is knowable. Real value; keep it.
- **Does not** block the case CLAUDE.md actually names — *"the governor must not be able to weaken its own governor"*, i.e. an agent landing a governance change unilaterally. The agent holds the credentials the check trusts.

The check is sound against the threat it can see. The gap is that its trust anchor and the actor it is meant to constrain are the same identity.

## Observable failure

An agent applies the label with no human involved; the gate reports ratified; the audit record shows owner attribution. Nothing downstream can tell that apart from a deliberate owner click. A human spot-checking the trail sees a ratification that never happened.

This is the #4671 shape: a check that cannot report a negative in the dimension it claims to measure.

## Mitigation used today

Provenance stated in a PR comment on each of #4697/#4698/#4704 — owner approved in-session, agent applied the label. That is a convention, not a control; the next agent need not follow it.

## Fix directions (none obviously right — wants a decision)

1. **Separate identity.** Agent uses its own token / GitHub App; `owner-ratified` only counts from the owner's account. Strongest, and the only one that makes attribution real. Cost: token plumbing, and the agent loses the ability to merge.
2. **Out-of-band signal.** Ratification requires something the agent cannot emit — a signed commit, a review from the owner's account under branch protection (`gh pr review --approve` is *also* blocked for one's own PR, which is useful here).
[...]

PR #4745, opened 2026-08-24
fix(governance): give the ratification backstop its missing producer

williamzujkowski:
**Main is red right now.** This fixes it. Found by an adversarial review of my own self-merged work, not by me.

## What broke

#4698 taught the ratification gate to require an *owner-applied* label, passing `RATIFICATION_LABEL_ACTOR` into the check. The pre-merge job produces it (`:143`). The post-merge backstop job **consumes it at `:232` and never produced it** — its evidence step wrote only `approvals` and `labels`.

An unset step output is an empty string, not an error. So the backstop read "applier cannot be established", correctly refused to call that ratified, and exited 1:

```
failure  b5948fdca  #4704
failure  d9b4cafe6  #4698
success  10bdacc8b  #4697   <- merged 4s before #4698 landed the workflow change
```

Every governor-path PR ratified by label reddens `main` after merge, and re-running cannot fix it.

The irony is worth stating plainly: this is a **consumer with no producer**, shipped inside the PR about attribution, on the same day I fixed roughly ten of that exact shape and wrote a memory note telling myself to trace producers. Fluency in a failure mode is not immunity to it.

## Second defect, same lines

`gh api --paginate --jq` applies the filter **per page**, so `| last` emits one line *per page* that contains a `labeled` event. Two such pages produce a bare second line in `$GITHUB_OUTPUT` → "Invalid format" step failure. The sibling `APPROVALS` query is guarded with `| tr '\n' ' '`; this one wasn't. Both timeline queries now end `| tail -n1`.

## The gate, not just the fix

`scripts/workflow-output-wiring.test.ts` asserts that every `steps.<id>.outputs.<name>` a job consumes has a producer in the same job.

**Mutation-verified against the real bug**, which is the only reason to trust it: it fails when run against the pre-fix `governor-review.yml` from `origin/main`, and passes against the fix. Not a synthetic mutant — the actual defect.

It also found a false positive on the first pass (`registry-refresh.yml` writes outputs via `appendFileSync` with a template literal rather than `echo`). Detection was broadened rather than the file excluded, since an exclusion list is how a gate quietly stops covering things. Steps that delegate to an action are skipped — their outputs aren't readable from YAML.

## Why the backstop is worth keeping strict

It refused to ratify on missing evidence, which is correct behaviour and exactly what the mission asks for — it failed *closed*. The bug was that it could never receive the evidence. Fixing the producer keeps the fail-closed semantics rather than loosening the gate to make the red go away.

## Ratification

Touches `.github/workflows/governor-review.yml` — a governance path, so this needs owner ratification and I am not self-merging it, despite it being a CI-red fix.

Note the bootstrap: once merged, the backstop runs from the merged tree, so the fixed workflow evaluates itself and self-heals.

### DIFF
--- a/.changeset/backstop-label-actor.md
+++ b/.changeset/backstop-label-actor.md
@@ -1,0 +1,26 @@
+---
+'nexus-agents': patch
+---
+
+fix(governance): the post-merge ratification backstop can see who applied the label
+
+#4698 taught the ratification gate to require an owner-applied label and pass
+`RATIFICATION_LABEL_ACTOR`. The pre-merge job produces it. The post-merge
+backstop job **consumes it and never produced it** — its own evidence step wrote
+only `approvals` and `labels`.
+
+An unset step output is an empty string, not an error, so the backstop read
+"applier unknown", correctly refused to treat that as ratified, and failed. Every
+governor-path PR ratified by label reddened `main` after merge, unfixable by
+re-running. #4698 and #4704 did exactly that.
+
+Adds the missing producer, and `| tail -n1` on both timeline queries: `gh api
+--paginate --jq` applies the filter per page, so `| last` can emit one line per
+page and corrupt the step output.
+
+Also adds `scripts/workflow-output-wiring.test.ts`, which asserts every
+`steps.<id>.outputs.<name>` a job consumes has a producer in the same job. It
+fails against the pre-fix workflow and passes after — the defect class is
+mechanically detectable and should not need a human to catch it twice.
+
+Governance path (the governor workflow) — requires owner ratification.
--- a/.github/workflows/governor-review.yml
+++ b/.github/workflows/governor-review.yml
@@ -143,7 +143,7 @@ jobs:
           LABEL_ACTOR=$(gh api "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/timeline" \
             --paginate \
             --jq '[.[] | select(.event == "labeled" and .label.name == "owner-ratified")] | last | .actor.login // empty' \
-            2>/dev/null || echo '')
+            2>/dev/null | tail -n1 || echo '')
           {
             echo 'files<<FILES_EOF'
             echo "${FILES}"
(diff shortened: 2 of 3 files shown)


williamzujkowski:
## Correcting my own framing on urgency

I have been describing main as "red" since this landed. That is true in a narrow sense and overstated in the way it matters, so here is the precise state:

```
governor-review runs on main (most recent first)
failure  b5948fdca  #4704   12:10:56Z
failure  d9b4cafe6  #4698   12:10:52Z
success  10bdacc8b  #4697   12:10:48Z   <- merged 4s before #4698 landed the change
success  7c8b22892          05:50:16Z
```

- The two failures are the **most recent** governor-review runs, and they cannot be re-run to green — so the workflow's status on main reads failed.
- **Every other workflow on main is green.** CI, CodeQL, Scorecard, Release, Documentation Gate, Link Check, Benchmark Extraction Gate — all passing on the last 12 runs.
- Nothing is broken for users, and non-governance merges are unaffected. Eight PRs have merged cleanly since.

What is actually broken: **the ratification backstop cannot pass**, so the next governor-path merge will fail the same way, and until then the post-merge check that verifies what landed is not providing coverage. That is a governance-instrument outage, not a build outage.

I should have said that from the start rather than repeating "main is red", which implies ongoing breakage and inflates the apparent urgency. The fix still matters — a governance gate that cannot pass is a gate nobody can rely on, and it will accumulate more permanently-failed commits — but it is not an emergency.

The PR is unchanged and still needs owner ratification: it touches `.github/workflows/governor-review.yml`, a governor-owned path, so I will not self-merge it regardless of how long it waits.

williamzujkowski:
## Ratification provenance

The owner approved the pending decisions in-session on 2026-08-24 ("I approve the pending decisions please proceed with them"), covering this PR and the #4759 release-version decision.

The `owner-ratified` label was applied by the **agent**, acting on that instruction. Recording it because the label alone cannot show it: the agent runs `gh` with the owner's credentials, so GitHub attributes the label to `williamzujkowski` either way. That gap is #4739.

The approval is genuine; the attribution is what would otherwise be misreported.

Scope check before merging: 3 files — the workflow fix, its changeset, and the wiring test. No other governor-owned path is touched.

PR #4757, opened 2026-08-24
chore(ci): derive the public API surface from the AST, not from name greps

### LINKED ISSUE #4749: ci: derive the public API surface from the AST — a name grep cannot answer a reachability question
Raised by the contrarian voter on the #4744 panel, and it is the most useful thing that vote produced. Quoting the substance: *"the author explicitly admits this is the third time today they have missed a structural API break because they rely on manual grepping... Manual grep is not a test."*

That is correct, and the three instances are on the record:

- **#4736** — `healthScore` widened to `number | null`, shipped as a patch.
- **#4740** — `VoteDecisionStatus` enum widening proposed as semver-minor; caught only because a dissenting voter challenged the category.
- **#4744** — claimed "not a public API change" after grepping for `ResultMetadata` by name in the barrels. Zero hits, and the conclusion was wrong: `TaskResult` is exported (`exports/core.ts:52`) and carries `metadata: ResultMetadata`, so the type is reachable **structurally**.

## Why the manual check keeps failing

A name grep answers "is this symbol re-exported?" The question that matters is "is this type reachable from the package entry point?" Those differ whenever a type is referenced by another exported type — which is the common case, not the edge case. There is no amount of care that makes a name grep answer a reachability question.

The failure is also silent and one-directional: it always reads as "not public", i.e. as permission to proceed.

## What to build

An AST-derived public API surface report, checked in and diffed in CI:

1. Generate the surface from `dist/index.d.ts` (API Extractor, or `tsc` + a small extractor). This resolves transitive reachability, which is exactly the part a grep cannot do.
2. Commit the report as a snapshot.
3. CI fails when a PR changes the snapshot without updating it, so the diff must be looked at deliberately.
4. Classify the diff: added symbol/optional field = minor; removed symbol, widened union in return position, required→optional on a read field = **major**, and the gate should say so rather than leaving it to the author's judgement.

Point 4 is the part that actually addresses the three failures. The break was visible in all three cases; what was missing was anything that forced the author to look.

## Existing tooling first

Per the repo's build-vs-buy gate: API Extractor (`@microsoft/api-extractor`) does 1-3 out of the box and is the obvious candidate. `tsd` and `are-the-types-wrong` cover adjacent ground. Step 4's severity classification is probably a thin layer on API Extractor's report rather than anything bespoke — check before building.

Note `scripts/check-new-unused-exports.ts` (#3024) already walks exports for a *different* question (unused ones). Worth checking whether that machinery can be reused rather than adding a parallel export walker — the anti-sprawl rule applies.

## Related

#4736, #4740, #4744, #4743. The immediate #4744 defect was avoided by taking the non-breaking route (panel chose it 5/6), so this issue is about the detection gap, not that specific fix.

williamzujkowski:
Closes #4749. Raised by the contrarian voter on the #4744 panel: *"Manual grep is not a test."*

## The mistake this prevents

Three type changes in one day, all from the same check:

| | claim | reality |
|---|---|---|
| #4736 | internal | `healthScore` widened to `number \| null`, shipped as a **patch** |
| #4740 | semver-minor | enum widening in a return position — **breaking** |
| #4744 | "not public" | `TaskResult` is exported and carries `metadata: ResultMetadata` |

Each time I grepped `src/exports/*.ts` for the symbol's **name**, found nothing, and concluded the change was internal.

A name grep answers *"is this symbol re-exported?"*. The question that decides semver is *"is this type reachable from the entry point?"* — and those differ whenever one exported type references another, which is the common case. The failure is also one-directional: it always reads as "not public", i.e. as permission to proceed. No amount of care fixes that; it is the wrong question.

## What this does

`extract-api-surface.ts` walks from `src/index.ts` and **follows type references transitively**, so a type that is public only through another type's signature is included.

That step is the whole point, and it is measurable: the export list alone yields 1912 symbols; following references yields **2225**. Those 313 types are public and invisible to a name grep.

`check-api-surface.ts` diffs against the committed `api-surface.txt` and fails on any change, printing the symbol and the exact member line that moved.

## The first version could not catch its own motivating case

Worth reporting rather than quietly fixing, because it is the defect class this whole series is about.

v1 diffed **line sets**. Running it against the real #4744 change reported *"Public API surface unchanged"* — because `tokensUsed?: number | undefined` already existed elsewhere in the file (under `StepResult`), so set membership said "present". A member line is only meaningful inside its declaration.

It now compares **per-symbol blocks**. Mutation-verified end to end against the actual #4744 edit: exit 1 with

```
REMOVED: InterfaceDeclaration ResultMetadata >  tokensUsed: number
ADDED:   InterfaceDeclaration ResultMetadata >  tokensUsed?: number | undefined
```

and exit 0 when reverted. I would not have trusted this gate without that check — and the first version would have passed a review that only read the code.

## Deliberate non-goals

**It does not classify severity.** It cannot know whether a removed symbol was load-bearing for a downstream consumer, and a gate that guessed would be trusted more than it deserves. It reports what moved and makes a human decide. The failure message states the rule of thumb (removed symbol / widened union / required→optional = breaking for readers) without pretending to apply it.

It also prints types as written rather than fully expanded, so deeply generic or conditional types are compared by their printed text.

## Cost

Uses `ts-morph`, already a dependency — no new tooling, per the build-vs-buy gate. Adds `pnpm api:surface` (regenerate) and `pnpm api:check` (verify), plus one CI step next to the existing producer/consumer gate.

`api-surface.txt` is 9423 lines. It is generated, marked do-not-edit, and its diff is the review surface — that is the feature.


williamzujkowski:
Self-merge note, flagged rather than left silent: this PR touched `.github/workflows/ci.yml`, which CODEOWNERS covers under **CI/CD workflows — supply chain protection** (line 14), requiring review.

It is *not* in the governance-of-the-governor section — that covers `src/audit/`, `src/governance/`, `scripts/inject-governance.ts`, `/governance/`, and `.github/workflows/governor-review.yml` specifically. The ratification gate did not trigger on this PR, which is the gate's own confirmation that `ci.yml` is not governor-owned.

So the self-merge was within the rules, but a supply-chain-sensitive path changed without a second reader. The change is one added CI step invoking a script in the same PR; worth a look if you are reviewing the batch.

Issue #4768, opened 2026-08-24
governance: the audit chain is empty in this deployment, and verify_audit_chain reports ok for it


williamzujkowski:
Found during the end-to-end validation CLAUDE.md asks for after behaviour-affecting fixes accumulate. It goes to the substrate's stated enabling condition, so I want the evidence on the record even though part of it is deployment-specific.

## 1. The audit chain has never written an event on this machine

```
$ find /home/william -maxdepth 6 -name "audit-*.jsonl"
(nothing)

$ verify_audit_chain /home/william/git/nexus-substrate/nexus-agents/.nexus-agents/audit
{ fileCount: 0, eventCount: 0, verification: { ok: true, eventCount: 0 } }

$ verify_audit_chain /home/william/.nexus-agents/audit
{ fileCount: 0, eventCount: 0, verification: { ok: true, eventCount: 0 } }
```

Not a broken data dir — seven other ledgers are actively written beside it: `capability-gaps.jsonl`, `tool-fitness/ledger.jsonl`, `usage/usage-2026-08.jsonl`, `ci-health/events.jsonl`, `token-ledger/ledger.jsonl`, `voting/correlations.jsonl`, `checkpoints/*.jsonl`.

The cause is not a bug. `initializeAuditLogger` (`cli-server-audit.ts:28-31`) returns `null` unless `securityConfig?.audit?.enabled === true`, and no config file in this repo sets it. Audit logging is **opt-in and off**.

## 2. The verifier cannot tell "clean" from "absent"

Both directories return `ok: true`. `verifyChain` short-circuits on `events.length === 0` (`audit-logger.ts:173`).

So the check that exists to inspect tamper-evidence reports success for a chain that does not exist. A human spot-checking sees green. Pointing the tool at the wrong directory produces the same green — and that specific mistake is easy: the repo-local and home data dirs both look plausible, and I checked both only because I had been bitten by it before.

This is the shape the mission text singles out: *prefer a gate that reports `unmeasured` over one that reports a default as a measurement*. `{ ok: true, eventCount: 0 }` is a default reported as a measurement.

## Why this matters more than the usual finding

CLAUDE.md's argument for full automation is explicit:

> Full automation is acceptable because everything is logged and a human can review it later — so the logs are load-bearing infrastructure, not exhaust.

[...]

PR #4773, opened 2026-08-25
fix(audit): say when a clean verdict verified nothing

williamzujkowski:
Implements the mitigation the threat model asks for by name. **Governor-owned path (`src/audit/`) — needs owner ratification; I have not self-merged.**

## The defect

`verifyChain` returns `{ ok: true }` in two cases where it checked no cryptographic link at all:

```ts
if (events.length === 0) return { ok: true, eventCount: 0 };                    // nothing to verify
if (events[0]?.hash === undefined) return { ok: true, eventCount: events.length }; // un-chained
```

Both verdicts are honest — nothing contradicts them — and both were indistinguishable from a genuinely verified chain.

Found during end-to-end validation: `verify_audit_chain` reported `ok: true` for **both** plausible data directories on this machine, because the audit chain has never written an event here (#4768). A human spot-checking tamper-evidence sees green. Pointing the tool at the wrong directory produces the identical result.

## The threat model already asked for this

T8 records **residual risk HIGH**:

> "OK" is ambiguous between "verified chained log" and "un-chained log, nothing to verify". **Mitigation would require the tool to report whether the log was chained at all** and to fail-closed (or at least warn loudly) on an un-chained log when chaining is expected by policy.

That was accurate when written. This adds the reporting half.

## What changes, and what deliberately does not

`ChainVerification`'s ok-variant gains `notVerified: 'empty' | 'unchained'`, absent when links were actually walked.

**`ok` is unchanged.** Two existing tests pin the empty case as accepted, one citing §1.4 explicitly — so I did not reverse a documented decision. The verdict simply stops implying assurance it never had.

**No caller behaviour changes.** Nothing that passed now fails. Whether `verify_audit_chain` should warn or fail-closed on `notVerified` is a policy decision, and stays with #4768.

## Tests

Four, all mutation-verified — removing each marker turns the corresponding test red:

- empty chain → `notVerified: 'empty'`
- un-chained log → `notVerified: 'unchained'` (T8's case)
- a genuinely verified chain → `notVerified` **absent**, so the field cannot become noise
- **through the registered MCP handler**, asserting the marker reaches the serialized response

The last one matters because the handler serializes the verdict whole, so the field flows through with no wiring — which is exactly the situation where I have previously shipped a field no reader ever sees. "The field exists" and "a caller sees it" are different claims, so the test exercises the handler rather than re-calling `verifyChain`.

## Verification

Full suite **28380 passed, 0 failed**. Audit + tool suites 309 passing. Typecheck 0, lint clean. API surface gate reports unchanged — `ChainVerification` is not publicly reachable.

### DIFF
--- a/.changeset/empty-chain-is-not-verified.md
+++ b/.changeset/empty-chain-is-not-verified.md
@@ -1,0 +1,37 @@
+---
+'nexus-agents': patch
+---
+
+fix(audit): verifyChain says when `ok: true` verified nothing
+
+`verifyChain` returns `{ ok: true }` in two cases where it checked no
+cryptographic link at all:
+
+- **zero events** — nothing to verify;
+- **an un-chained log** — the first event carries no `hash`, so the whole batch
+  short-circuits and no links are walked.
+
+Both are honest verdicts (nothing contradicts them) and both were
+indistinguishable from a genuinely verified chain. `verify_audit_chain` reported
+`ok: true` for an empty directory, so pointing it at the wrong path — or at a
+deployment where audit logging is simply off (#4768) — looked identical to a
+clean chain.
+
+The ok-variant now carries `notVerified: 'empty' | 'unchained'`, absent when
+links were actually checked. `ok` is unchanged, so the threat model's accepted
+position on empty logs stands; what changes is that the verdict no longer
+implies assurance it does not have.
+
+This is the mitigation the threat model asks for by name. T8 records **residual
+risk HIGH** — _"'OK' is ambiguous between 'verified chained log' and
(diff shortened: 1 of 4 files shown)


williamzujkowski:
## Ratification provenance

Owner approved in-session on 2026-08-25 ("I approve 4773"). The `owner-ratified` label was applied by the **agent** on that instruction — recorded because the label cannot show it: `gh` runs with the owner's credentials, so GitHub attributes it to `williamzujkowski` either way (#4739).

Scope check before merge: 3 files — `audit-logger.ts`, its chain-verify tests, and the MCP tool test. No other governor-owned path touched.

Issue #4772, opened 2026-08-25
fix(pipeline): a dry run whose planner failed returns the same envelope as one that succeeded


williamzujkowski:
Found by running `run_dev_pipeline --dryRun` twice against a current build (4.0.0, restarted MCP server) as part of the end-to-end validation. One run planned well; one run's planner failed outright. **Their result envelopes are identical.**

## 1. A failed dry run is indistinguishable from a successful one

Both runs returned:

```
completed: false   securityPassed: false   voteIterations: 1   qaIterations: 0   tasks: []
```

Run A produced a detailed implementation plan — options table, an ADR on whether the check should gate, red/green step order.

Run B produced no plan at all. Its `plan` field contains the prompt echoed back plus the research/outcome preamble, and the outcome context carried `parse: Model returned empty response` five times.

Nothing in the returned shape separates these. A caller — or an agent looping on this tool — cannot tell "here is your plan" from "the planner returned nothing five times". The only signal is whether the `plan` string happens to look like a plan, which is not a check anything can perform.

## 2. `securityPassed: false` on a dry run reports a stage that never ran

`dev-pipeline.ts:697`, the dry-run return path:

```ts
return {
  completed: false,
  plan: planResult.plan,
  tasks: [],
  voteIterations: planResult.iterations,
  qaIterations: 0,
  securityPassed: false,   // security never executed
};
```

A dry run stops after plan+vote by design (`:314`). Reporting `securityPassed: false` states that the security gate did not pass, when it was never invoked — absence rendered as a negative verdict. `securityPassed: boolean` cannot express "not run".

## What I checked and found sound

- **Dry runs do NOT pollute the learning loop.** I suspected they might, because `recordPipelineHindsight` computes `outcomeMatched: result.completed && result.securityPassed` (`:509`), which is permanently false for a dry run. But the dry-run path returns at `:316`, before the recorder is reached. No hindsight record is written. Worth stating explicitly since the arithmetic looks alarming in isolation.
- The "Quality trend: DECLINING" line in run B's context reflects pre-existing recorded outcomes, not these runs.

## Suggested

1. **Make the dry-run result say what it is.** Either a discriminated result (`phase: 'planned'`) or `securityPassed?: boolean` left absent, so "not run" is representable. Same fix shape as `tokensMeasured` (#4744) and `unmeasuredSteps` (#4734).
[...]

PR #4792, opened 2026-08-25
fix(mcp): say why the dev pipeline did not complete, and keep the verdict

### LINKED ISSUE #4789: run collapses the dev-pipeline result to a generic error exactly when securityRan/planStatus matter
## Finding

`run` delegates to the dev pipeline: `run-tool.ts:285` maps strategy `'dev-pipeline'` to `runDevPipelineForGoal`, which returns the full `DevPipelineResult` (`dev-pipeline-tool.ts:254-266`).

On success, `run-tool.ts:451` serialises `exec` wholesale, so `securityRan` and `planStatus` reach the caller.

On failure they are thrown away. `run-tool.ts:408-410`:

```ts
if (record['completed'] === false) {
  return 'Engine reported failure: the dev pipeline did not complete';
}
```

which flows to `run-tool.ts:450` → `toolStructuredError({ errorCategory: 'business', message: engineFailure })`, discarding `exec.result` entirely.

`completed === false` is **exactly** the condition under which the two fields carry information.

## Observable failure

A dry run (`dev-pipeline.ts:721-731` → `completed: false, securityRan: false, planStatus: 'empty'`) or a harness-mode run (`dev-pipeline.ts:830-838`) invoked through `run` returns only:

> Engine reported failure: the dev pipeline did not complete

The caller cannot distinguish:
- "the security gate rejected this" from "the gate never ran", or
- "the planner produced nothing" from "the plan was fine, implementation was skipped".

That is precisely the confusion #4772 was filed to fix. Direct `run_dev_pipeline` callers do see the fields (fixed in #4777); `run` callers do not — and `run` is the documented default entry point.

Worse than a missing field: a dry run is a *normal, successful* use of the pipeline, and `run` reports it as an engine failure.

## Fix sketch

Preserve the structured result on the `completed: false` path — either attach `exec.result` to the error envelope, or stop treating `completed: false` as an engine failure when the run was a dry run / harness mode, which are expected non-completions rather than faults.

## Provenance

Found by an adversarial review of #4777, then independently traced through `run-tool.ts` from the strategy mapping to both return paths. Two sibling claims from the same review were checked and **refuted** (`workflows/index.ts` "orphan" exports are dead-but-valid public API with test importers, no observable failure), so this one is not a blanket acceptance of that review.

williamzujkowski:
Fixes #4789.

## Problem

`run-tool.ts` treated every dev-pipeline non-completion identically:

```ts
if (record['completed'] === false) {
  return 'Engine reported failure: the dev pipeline did not complete';
}
```

and at the call site discarded `exec.result` entirely. So a `run` caller could not distinguish **"the security gate rejected your change"** from **"the security gate never ran"** — exactly the distinction #4772 → #4783 put into `securityRan` one layer down, thrown away one layer up.

## Correction to the original report

The issue as filed cited dry-run and harness mode. **Those cannot occur on the `run` path** — `runDevPipelineForGoal` parses `{ task: goal }` only, so `dryRun` stays at its `false` default and harness mode is never selected. I verified this before building on it and corrected the issue.

The finding survives with a narrower, real failure. The two reachable non-completions are:

| Case | `securityRan` | Old message |
|---|---|---|
| scan ran, rejected the change | `true` | *did not complete* |
| quality gate short-circuited first | `false` | *did not complete* |

## Fix

`describeIncompletePipeline` names the reason — empty plan, security rejected, or stopped before the gate — and the engine's result travels in the error envelope's `detail`.

`securityRan` is read as **three-valued on purpose**: absent means a producer predating the field, *not* `false`, so the generic message stands rather than claiming a reason we don't have. That is the distinction the mutation table below tests.

## Verification

Red/green — four tests written first, three failed against `main`:

```
× says the security gate REJECTED when the scan actually ran
× says the run stopped BEFORE the security gate when it never ran
× names an empty plan as the reason rather than a generic non-completion
  Tests  3 failed | 31 passed (34)
```

After: `34 passed`. Mutation-verified:

| Mutation | Result |
|---|---|
| back to one generic message, no `detail` | 3 failed |
| absent `securityRan` treated as `false` | 1 failed |

The fourth test is the one that catches the second mutation — a pre-#4783 producer must not be reported as "the gate never ran".

`tsc` clean, API surface unchanged.

## Deliberately not in scope

`detectEngineFailure` already carves out consensus: *"a `rejected` outcome is the verdict the caller asked for, not an engine fault."* A security scan that correctly rejects unsafe code is arguably the same shape — the pipeline **worked**, and its answer was no. Reclassifying that away from `errorCategory: 'business'` is a wider behaviour change than this fix needs, so it stays open as a separate question on #4789 rather than being decided here.

### DIFF
--- a/.changeset/run-preserve-pipeline-verdict.md
+++ b/.changeset/run-preserve-pipeline-verdict.md
@@ -1,0 +1,18 @@
+---
+'nexus-agents': patch
+---
+
+say why the dev pipeline did not complete on the `run` path, and keep the result
+
+Every non-completion reaching `run` collapsed to "the dev pipeline did not
+complete", and `exec.result` was discarded — so a caller could not tell a
+security gate that REJECTED the change from one that never ran, which is the
+distinction #4772/#4783 added `securityRan` to provide.
+
+The message now names the reason (empty plan, security rejected, stopped before
+the gate) and the engine's result travels in the error envelope's `detail`.
+`securityRan` is read as three-valued: absent means a producer predating the
+field, not `false`, so the generic message stands rather than claiming a reason
+we do not have.
+
+Fixes #4789.
--- a/packages/nexus-agents/src/mcp/tools/run-tool.test.ts
+++ b/packages/nexus-agents/src/mcp/tools/run-tool.test.ts
@@ -9,15 +9,31 @@ import { join } from 'node:path';
 // #3712: capture the trustTier handed to runDevPipelineForGoal through the
 // run → dev-pipeline executor path (the "hole" — a real RequestContext that ran
 // a real research stage on a possibly-untrusted goal with an absent tier).
-const runDevPipelineForGoalMock = vi.fn((_goal: string, _trustTier?: string) =>
-  Promise.resolve({
-    completed: true,
-    plan: 'plan',
-    tasks: [],
-    voteIterations: 1,
-    qaIterations: 1,
-    securityPassed: true,
-  })
+/**
+ * The pipeline result shape the mock returns. Declared rather than inferred so
+ * a test can set the #4783 provenance fields — inference from the happy-path
+ * literal alone rejects `securityRan`/`planStatus` as unknown properties.
+ */
+interface FakeDevPipelineResult {
+  completed: boolean;
+  plan: string;
+  tasks: never[];
+  voteIterations: number;
+  qaIterations: number;
+  securityPassed: boolean;
+  securityRan?: boolean;
+  planStatus?: 'empty';
+}
+const runDevPipelineForGoalMock = vi.fn(
+  (_goal: string, _trustTier?: string): Promise<FakeDevPipelineResult> =>
+    Promise.resolve({
+      completed: true,
+      plan: 'plan',
+      tasks: [],
+      voteIterations: 1,
+      qaIterations: 1,
+      securityPassed: true,
+    })
 );
 vi.mock('./dev-pipeline-tool.js', () => ({
   runDevPipelineForGoal: (goal: string, trustTier?: string) =>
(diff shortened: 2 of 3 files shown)


williamzujkowski:
## Correction: my PR description overstated which branches can fire

#4792's description said the two reachable non-completions on the `run` path are "scan ran and rejected" and "quality gate short-circuited in `blocking` mode". I re-checked, and **only the first is reachable**.

`runDevPipelineForGoal` (`dev-pipeline-tool.ts:258`) parses `DevPipelineInputSchema.parse({ task: goal })` and nothing else, so every option takes its default: `dryRun: false`, `mode: 'autonomous'`, `qualityGate: 'off'`. That makes all three of the other producers unreachable through `run`:

| Branch in `describeIncompletePipeline` | Producer | Reachable via `run`? |
|---|---|---|
| `planStatus === 'empty'` | `buildDryRunResult` only | **no** — dry runs cannot be requested |
| `securityRan === false` | dry run / harness mode / `qualityGate: 'blocking'` | **no** — none selectable |
| `securityRan === true` | post-scan return | yes |

`detectEngineFailure` has exactly one caller (`run-tool.ts:473`), so "unreachable via `run`" means unreachable, full stop. Two of the three branches I added cannot fire.

The fix is still a real improvement — the one case that *does* occur now names itself instead of saying "did not complete", and the verdict travels in `detail`. But I shipped two branches that cannot fire while fixing a defect of exactly that class, which is worth recording rather than leaving in a PR description that reads as if all three were live.

Filed as a follow-up: either widen `run`'s input so those modes become expressible (making the branches real), or trim them. That is a product decision about `run`'s surface, not a cleanup.

Issue #4838, opened 2026-08-25
system-review reports 'No vulnerabilities found' when pnpm audit fails — the failure path is the detection path


williamzujkowski:
## Finding

`system-review`'s security phase reports **"✓ No vulnerabilities found"** in precisely the case where vulnerabilities exist.

The chain, each link read against main at v4.3.4:

1. `runPhase4` (`system-review.ts:150-156`) runs `safeExec('pnpm audit --json')`.
2. `safeExec` → `safeExecSandboxed` → `execSync`, which **throws on a non-zero exit**; the catch (`sandbox-exec.ts:183-189`) returns `null`.
3. **`pnpm audit` exits non-zero when it finds vulnerabilities.** So the failure path is the *detection* path.
4. `null` → `return { ...def, parseError: true }`, and `def` is `{ totalVulns: 0, high: 0, moderate: 0, low: 0 }`.
5. `printPhase4` (`system-review-helpers.ts:107`) branches **only** on `totalVulns === 0` → prints the pass line.
6. `calculateHealthScore:49-50` deducts `high * PENALTY + moderate * PENALTY` — both zero, so **no penalty is applied**.

A vulnerable repo therefore gets a green security phase and an unpenalised health score, and the `--create-issue` body tables High: 0 / Moderate: 0 / Low: 0.

## Disclosure: partial, and not where it counts

`parseError` **is** set, and `getSecurityItems` (`:211`) does emit `WARNING: Security audit parsing failed - verify pnpm audit manually`. That is real, and it is why this is not a total blind spot.

But it appears only in a later action-items section. The three places a reader actually forms the judgement — the Phase 4 verdict line, the health score, and the filed issue — all ignore `parseError` entirely. The signal exists in the data and is dropped by every consumer that matters.

## Why this one is worth prioritising

The failure mode is inverted, not merely silent. A tool that says nothing when it cannot check is a gap; one that says **"no vulnerabilities"** when the check failed is worse than not running it, because it manufactures the assurance a reviewer relies on. This repo's own doctrine names that as the p1 case: a record that misreports is worse than a missing one.

It also compounds — `parseError: true` is indistinguishable from a genuine clean audit in the issue body, so the artifact that outlives the terminal session carries the wrong answer with no caveat.

## Fix

[...]

PR #4847, opened 2026-08-25
fix(release): let the secret scan report that it did not run

### LINKED ISSUE #4839: Release secret-scan reports clean when git fails — the pipeline hides the exit code the catch relies on
## Finding

`validateSecurity`'s hardcoded-secret scan (`release-validate-helpers.ts:68-97`) cannot detect its own failure for the case it was written to detect.

```ts
execSync('git diff HEAD~10..HEAD -- "*.ts" "*.js" | grep -iE "(api[_-]?key|secret|password|token)" | head -5', …)
```

A shell pipeline's exit status is the **last** command's. `head` succeeds essentially always, so a `git diff` failure never propagates.

**Reproduced** in a fresh repo with one commit:

```
git diff HEAD~10..HEAD -- "*.ts"            → exit 128
git diff HEAD~10..HEAD -- "*.ts" | grep … | head -5   → exit 0, stdout empty
```

So on a shallow clone or a branch with fewer than ten commits, `execSync` does not throw, `result.trim()` is empty, no finding is pushed, and the scan silently reports clean.

## The catch cannot fire for the case its own remediation names

`:86-97` records a `Secret scan did not run` finding and says:

> *"reaching this catch means the scan itself failed (git error, timeout), not that the tree is clean"*
> remediation: *"Re-run with a full git history available, then review the output."*

**A git error does not reach it** — the pipe swallows it. Only a timeout does. And "full git history available" is precisely the shallow-clone scenario the pipe masks.

## Consequence

`:104` computes `passed: !anyOf(findings, f => f.severity === 'error', false)`, with a comment stating:

> *"whenEmpty = false: with the failure above recorded as a finding, an empty list now genuinely means 'scanned, nothing found' (#4581)."*

That reasoning is sound **and its premise does not hold**. The failure above is not recorded for the git case, so an empty list can still mean "did not scan". Release validation reports the security expert as passed having scanned nothing.

#4581 correctly identified the empty-list hazard and fixed it at the wrong layer: it hardened the verdict while leaving the detector unable to report the failure the verdict depends on.

## Fix

Stop relying on the pipeline's status. Run `git diff` on its own, check its exit code explicitly, and filter in JS — or add `set -o pipefail` via an explicit shell invocation. Either makes the existing `Secret scan did not run` branch reachable for the case it describes; no other change is needed, because the handling is already correct.

**Disclosure: none, and the adjacent comments assert the opposite** — both the catch comment and the `whenEmpty` comment tell a reader this case is covered.

## Provenance

From a calibrated sweep of `src/cli`. I reproduced the pipeline-masking behaviour directly rather than reasoning about it; the exit codes above are from an actual run. Same shape as #4838 — a clean verdict produced by a check that did not run — and both sit on paths that gate a release.

williamzujkowski:
Fixes #4839.

## The defect

`validateSecurity`'s hardcoded-secret scan ran this:

```
git diff HEAD~10..HEAD -- "*.ts" "*.js" | grep -iE "(api[_-]?key|secret|password|token)" | head -5
```

A shell pipeline's exit status is its **last** command's. `head` succeeds essentially always, so `git`'s exit code never reached `execSync`. On a shallow clone — or any branch with fewer than ten commits — the command exits **0 with empty output**, which is byte-for-byte what a clean tree produces.

Reproduced directly rather than reasoned about, in a fresh repo with one commit:

```
git diff HEAD~10..HEAD -- "*.ts"                       → exit 128
git diff HEAD~10..HEAD -- "*.ts" | grep … | head -5    → exit 0, stdout empty
```

## Why the existing handling didn't save it

The catch block already recorded a `Secret scan did not run` finding, and its remediation read *"re-run with a full git history available"*. **A missing git history was exactly the case that could not reach it.** Only a timeout did. #4581 had correctly identified the empty-list hazard and hardened the verdict — one layer above a detector that still could not report the failure the verdict depends on.

Net effect: release validation reported the security expert as **passed**, having scanned nothing.

## The fix

`scanRecentCommitsForSecrets` runs `git diff` on its own and filters in JS, returning either the matches or the reason it could not scan. The existing finding becomes reachable for the case it describes, and the git error is now carried into its description — previously the operator was told the scan failed but not why.

## Testing

The old `#4581` test asserted this catch by making a mocked `execSync` throw for `git diff` — something the real command could not do. **A mock cannot reproduce shell exit-status semantics**, so the new tests use the real `execSync` against a real temporary git repo with a single commit. That is the actual bug, exercised.

Mutation-verified: restoring the pipe fails the did-not-run test and leaves the other two green — the test pins this defect specifically, not the function generally.

24 tests green across both suites; `tsc` and eslint clean.

## Scope

`validateSecurity`'s two other checks were examined in passing and are sound: `npm audit` runs unpiped, and the `.env` check is a filesystem stat. Not changed. #4838 is the same shape in `system-review` and remains open.

### DIFF
--- a/.changeset/secret-scan-reports-git-failure.md
+++ b/.changeset/secret-scan-reports-git-failure.md
@@ -1,0 +1,22 @@
+---
+'nexus-agents': patch
+---
+
+let the release secret scan report that it did not run
+
+The hardcoded-secret scan in `validateSecurity` piped `git diff` through
+`grep | head`. A shell pipeline's exit status is its _last_ command's, and
+`head` succeeds essentially always — so a `git diff` failure (a shallow clone,
+or any branch with fewer than ten commits) exited 0 with empty output, which is
+indistinguishable from a clean tree.
+
+The catch block that records `Secret scan did not run` therefore could not fire
+for the case its own remediation names ("re-run with a full git history
+available"); only a timeout reached it. Release validation reported the security
+expert as passed having scanned nothing.
+
+`scanRecentCommitsForSecrets` now runs `git diff` on its own and filters in JS,
+returning either the matches or the reason it could not scan, and the git error
+is carried into the finding's description.
+
+Fixes #4839.
--- a/packages/nexus-agents/src/cli/release-secret-scan.ts
+++ b/packages/nexus-agents/src/cli/release-secret-scan.ts
@@ -1,0 +1,58 @@
+/**
+ * Recent-commit secret scan for release validation.
+ *
+ * Separate from `release-validate-helpers` because it is the one check there
+ * that shells out to git, and the distinction it has to preserve — "scanned,
+ * nothing found" versus "did not scan" — is worth testing against a real
+ * repository rather than a mocked `execSync` (#4839).
+ *
+ * @module cli/release-secret-scan
+ */
+
+import { execSync } from 'node:child_process';
+import { CLI_SUBPROCESS_TIMEOUTS } from '../config/timeouts.js';
+
+/** Outcome of the recent-commit secret scan: it either ran, or it did not. */
+export type SecretScanResult =
+  | { readonly ok: true; readonly matches: readonly string[] }
+  | { readonly ok: false; readonly reason: string };
+
+/** Options for {@link scanRecentCommitsForSecrets}; both exist for testing. */
+export interface SecretScanOptions {
+  readonly cwd?: string;
+  readonly range?: string;
+}
+
+const SECRET_LINE_PATTERN = /api[_-]?key|secret|password|token/i;
+const MAX_REPORTED_MATCHES = 5;
+
+/**
+ * Scan recent commits for secret-like tokens, distinguishing "scanned, nothing
+ * found" from "did not scan".
+ *
+ * This runs `git diff` on its own and filters in JS. It previously piped git
+ * through `grep | head`, and a shell pipeline's exit status is its *last*
+ * command's: `head` succeeds essentially always, so a git failure — a shallow
+ * clone, or any branch with fewer than ten commits — exited 0 with empty
+ * output and was indistinguishable from a clean tree (#4839).
+ */
+export function scanRecentCommitsForSecrets(options: SecretScanOptions = {}): SecretScanResult {
+  const range = options.range ?? 'HEAD~10..HEAD';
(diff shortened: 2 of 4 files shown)


williamzujkowski:
The Producer/Consumer gate was correct and I was wrong about the shape of the fix.

`scanRecentCommitsForSecrets` had a production consumer — `validateSecurity`, in the same file — but the gate counts *cross-file* imports, and by that standard the `export` keyword itself had no consumer. It existed only so the test could reach the function. That is exactly what the ratchet is for.

The two escapes the failure message offers both looked wrong here: there was nothing to "wire up", and `@export-no-consumer-yet` asserts a production consumer is still to come, which would have been false. Reaching for the marker anyway is how a gate stops meaning anything — the script's own comments say so.

So instead the scanner moved to its own module, `release-secret-scan.ts`, and `release-validate-helpers.ts` imports it. The consumer is now real and cross-file, the gate passes on its own terms, and the separation is defensible independently: it is the one check in that file that shells out to git, which is precisely why it needed a test against a real repository.

Ratchet run locally on the amended branch:

```
Producer/consumer check: 1 new file(s) have production consumers — OK.
```

24 tests still green across both suites; `tsc` and eslint clean. No change to any gate.

williamzujkowski:
CI caught my test making the exact assumption this PR is about.

`Test (ubuntu-latest)` failed on **`reports a clean scan as scanned, not as failed`** — the control case, not the fix. It scanned the checkout with `HEAD~1..HEAD`, and `actions/checkout` clones at depth 1, so `HEAD~1` does not resolve on the runner. Confirmed locally:

```
git clone --depth 1 … && git log --oneline | wc -l   →  1
```

So the scan correctly returned `ok: false` and the test — which asserted `ok: true` — failed. The production code did the right thing; the assertion depended on a full git history it was never guaranteed.

Worth noting because it is the same defect one level up: I wrote a test whose result depended on the environment rather than on the behaviour, and a shallow clone is precisely the case the fix exists for. Had the pipe still been masking git's exit code, this test would have passed on the runner and told me nothing.

It now builds its own two-commit fixture, like the other two cases already did, and additionally asserts `matches` is empty — otherwise "scanned successfully and found a secret in a file with no secret" would satisfy it.

3 tests green locally; lint clean. No production change in this commit.

Issue #4854, opened 2026-08-25
the collaboration-session aggregator performs no conflict detection, so conflicts is always empty


williamzujkowski:
Split out of #4831 because it is separate work, filed now rather than when someone gets to it.

`buildAggregatedResult` (`agents/collaboration/session-helpers.ts`) returns `conflicts: []` unconditionally. #4831 changed `conflictCount` to be *derived* from that list rather than restated as a literal `0`, so the two can no longer disagree — but the list itself is still always empty, so the count is still always zero. Deriving it was drift-prevention, not detection.

The sibling has the machinery: `ResultAggregator.applyStrategy` → `mergeObjects` (`aggregator-helpers.ts:187`) compares outputs field-by-field and emits a `ResultConflict` per disagreement, via `handleFieldConflict` (`:164`). The session path never calls it, so two experts returning contradictory outputs for the same key produce a session that reports zero conflicts.

**Why it matters:** `AggregationMetadata.conflictCount` is public API (`api-surface.txt`), and a consumer reading zero conflicts from a genuinely conflicted session gets the same answer as from a clean one. Same shape as the rest of the audit series — a count that cannot be non-zero is not a measurement.

**What would change:** run the existing conflict detection over `results` in `buildAggregatedResult`, or route this path through `ResultAggregator` so there is one implementation rather than two. The second is preferable on DRY grounds — two producers of `AggregatedResult` is what caused #4831 — but it is a larger change: the session path takes `TaskResult[]` where the aggregator takes `ExpertResult[]`, and only the latter carries `confidence` and `expertId`.

**Trigger to pick this up:** either a consumer needs real conflict counts from collaboration sessions, or the `TaskResult`/`ExpertResult` split gets reconciled — at which point routing both producers through `ResultAggregator` becomes cheap and closes #4831's root cause too.

Related: #4831 (the hardcoded metadata fields), #4743 (the token-usage half of the same drift).

williamzujkowski:
Trigger check before picking this up — **not met, leaving deferred.**

This issue's own unblock condition is "either a consumer needs real conflict counts from collaboration sessions, or the `TaskResult`/`ExpertResult` split gets reconciled". Both still hold as written:

- `buildAggregatedResult` still returns `conflicts: []` unconditionally, so the defect is unchanged and STILL OPEN.
[...]

PR #4855, opened 2026-08-25
fix(collaboration): disclose that session confidence is unmeasured

### LINKED ISSUE #4831: buildAggregatedResult hardcodes conflictCount 0 and averageConfidence 1.0 while its sibling computes both
## Verified: `session-helpers.buildAggregatedResult` hardcodes two fields its sibling computes

Second finding from the calibrated sweep of `src/adapters` / `src/agents/collaboration` / `src/cli-adapters/routing`. Verified against main at v4.3.4. (The two security findings from that sweep are recorded privately per the discovered-issues protocol, not here.)

## Finding

Two producers implement the same `AggregatedResult` shape, and only one measures.

`session-helpers.ts:283-296` — live via `collaboration-session-helpers.ts:143`:

```ts
conflicts: [] as ResultConflict[],
metadata: {
  resultCount: results.length,
  conflictCount: 0,          // hardcoded
  averageConfidence: 1.0,    // hardcoded — a perfect score
  ...
}
```

`result-aggregator.ts:145-153` — the sibling:

```ts
conflicts,
metadata: {
  conflictCount: conflicts.length,
  averageConfidence: Math.round(avgConfidence * 100) / 100,
  ...
}
```

Same interface, two producers, one of which reports a perfect, conflict-free session unconditionally.

## What makes this worth reporting rather than a nit

**It is public API.** `AggregationMetadata` is in `api-surface.txt:401-406` with both `averageConfidence` and `conflictCount` as required fields, so external consumers read them.

**The drift-prevention effort stopped one line short.** The comment immediately below the hardcodes reads:

> `#4743: shared with result-aggregator so the two cannot drift — they were duplicate expressions computing the same thing.`

Someone recognised the two-producer drift risk, extracted `summarizeTokenUsage` to fix it for token usage — and left `conflictCount: 0` and `averageConfidence: 1.0` hardcoded directly above the comment saying the two must not drift.

**Disclosure check:** partial, and not for these fields. `qualityScore` in the same object *is* genuinely computed and does vary — but a real quality score tells a consumer nothing about whether `conflictCount` was measured. A degraded session with contradictory experts is byte-identical to a clean one on both hardcoded fields. That is the `memory_stats` shape (#4828), not the disclosed shape.

## Fix

Compute both in `buildAggregatedResult`, or delegate to the sibling — the extraction pattern `summarizeTokenUsage` already established one line below is exactly the precedent.

Not implementing: the two producers exist for a reason I have not traced, and collapsing them is a design call rather than a mechanical fix. Worth pairing with whoever picks up #4743, since that issue's own extraction is what makes the omission visible.

williamzujkowski:
Fixes #4831. Splits #4854 out.

## What the issue said, and what turned out to be true

The report was: two producers implement `AggregatedResult`, and `session-helpers.buildAggregatedResult` hardcodes `conflictCount: 0` and `averageConfidence: 1.0` where `ResultAggregator.buildResult` computes both. That framing suggests copying the sibling's expressions across.

Tracing the inputs says otherwise. **No input reaching the session builder carries a confidence signal at all** — `TaskResult`, `ResultMetadata`, `VoteMessage` and `ExpertParticipation` all lack the field. The aggregator computes from `ExpertResult.confidence`, which this path never has. There was nothing to copy.

So `1.0` was not a lazy restatement of an available value; it was a **placeholder presented as a perfect score**, which is the worse of the two.

## The fix follows a precedent already in this interface

`AggregationMetadata` already carries `unmeasuredResults` (#4743) for exactly this situation on token counts, and `ResultMetadata.tokensMeasured` (#4734) is the same pattern one level down: a required numeric field that may be a placeholder, disclosed by an adjacent optional boolean. Adding a third instance of an established shape beats inventing a fourth.

`confidenceMeasured?: boolean` — optional, so no existing producer or consumer breaks. The session builder reports `0` / `false`; the aggregator reports its computed value / `true`.

**The placeholder moves from 1.0 to 0 deliberately.** It fails in the safe direction: a consumer thresholding on confidence previously passed unconditionally on this path. No in-tree consumer reads the field — `api-surface.txt` lists it as public API, so external ones may.

## conflictCount is honestly a smaller thing than the issue implies

`conflictCount: 0` was *consistent* with the `conflicts: []` on the line above it — not a contradiction within the object. It is now derived from that list so the two cannot drift, but that is drift-prevention, not detection: **the list is still always empty, because this path performs no conflict detection.** Wiring the existing `mergeObjects` detection in, or routing both producers through `ResultAggregator`, is #4854 — filed now rather than left as a comment.

Calling that fixed here would be the same class of misreport the change is about.

## Ruled out while verifying

`ResultAggregator.buildResult` divides by `results.length`, which looked like a NaN on empty input. It is not reachable: `aggregate` rejects an empty result list at `:73` before `buildResult` is called. Dropped rather than "fixed".

## Testing

Three tests, red first; each of the three production hunks reverted individually fails exactly one. Full collaboration suite: **979 tests, 38 files, green.** `tsc` and eslint clean.

### DIFF
--- a/.changeset/aggregation-confidence-measured.md
+++ b/.changeset/aggregation-confidence-measured.md
@@ -1,0 +1,28 @@
+---
+'nexus-agents': patch
+---
+
+stop reporting a perfect confidence for a session that never measured one
+
+`buildAggregatedResult` returned `averageConfidence: 1.0` — the best possible
+score — for every collaboration session, and `conflictCount: 0` as a literal.
+Its sibling `ResultAggregator.buildResult` computes both from the same
+`AggregatedResult` shape.
+
+The two are not symmetric, though, and the asymmetry is the point: **no input
+reaching the session builder carries a confidence signal at all.** Neither
+`TaskResult`, `ResultMetadata`, `VoteMessage` nor `ExpertParticipation` has the
+field; the aggregator path computes it from `ExpertResult.confidence`. So there
+was nothing to compute, and the value was a placeholder presented as a score.
+
+Following the precedent already in this interface (`unmeasuredResults`, #4743,
+and `ResultMetadata.tokensMeasured`, #4734), `AggregationMetadata` gains an
+optional `confidenceMeasured`. The session builder reports `0` with
+`confidenceMeasured: false`; the aggregator reports its computed value with
+`confidenceMeasured: true`. The placeholder now fails in the safe direction — a
+consumer thresholding on confidence previously passed unconditionally.
+
+`conflictCount` is now derived from the `conflicts` list so the two cannot
+drift. That path still performs no conflict detection, which is #4854.
+
+Fixes #4831.
--- a/.changeset/aggregation-confidence-measured.md
+++ b/.changeset/aggregation-confidence-measured.md
@@ -1,5 +1,5 @@
 ---
-'nexus-agents': patch
+'nexus-agents': minor
 ---
 
 stop reporting a perfect confidence for a session that never measured one
@@ -21,6 +21,9 @@ optional `confidenceMeasured`. The session builder reports `0` with
 `confidenceMeasured: false`; the aggregator reports its computed value with
 `confidenceMeasured: true`. The placeholder now fails in the safe direction — a
 consumer thresholding on confidence previously passed unconditionally.
+
+`AggregationMetadata` gaining an optional field is an additive public-API
+change, so this is a minor rather than a patch.
 
 `conflictCount` is now derived from the `conflicts` list so the two cannot
 drift. That path still performs no conflict detection, which is #4854.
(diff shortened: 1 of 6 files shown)


williamzujkowski:
The API-surface gate was right and my semver level was wrong.

`AggregationMetadata` gaining `confidenceMeasured?: boolean` is a public-API change, and the gate's own guidance settles the classification: *"Additive optional fields and new symbols are minor."* I had marked the changeset `patch`.

Worth being precise about why it is minor and not breaking. The field is **optional**, so no existing producer of `AggregationMetadata` stops compiling and no existing reader loses anything. What changes is that readers gain a signal they did not have. A *required* field becoming optional would be the breaking direction, and that is not what happened here.

Both fixed in the follow-up commit: `pnpm api:surface` regenerated, and the changeset bumped to `minor` with a line saying why.

This repo has been bitten by the other answer — #4736 is `healthScore` widened to `number | null` and shipped as a patch, a breaking change that skipped the gate entirely. Getting it wrong in the safe direction still means the release notes understate what changed, which is what the gate exists to stop.

Issue #4802, opened 2026-08-25
Branch protection enforces neither the governor ratification gate nor CODEOWNERS


williamzujkowski:
**Owner-gated — changing branch protection is your call, not mine. Filing, not acting.**

## What I found

```
$ gh api repos/nexus-substrate/nexus-agents/branches/main/protection
contexts:                       ["CI Success"]
require_code_owner_reviews:     false
required_approving_review_count: 1
enforce_admins:                 false
```

Two independent gaps, both on the governor path.

### 1. Only `CI Success` can block a merge

Eleven PR-triggered workflows exist; `CI Success` aggregates jobs from **`ci.yml` alone**. Every job in every other workflow is advisory, whatever its result:

| Workflow | Jobs that cannot block |
|---|---|
| `governor-review.yml` | `governor-review`, **`governor-ratification`**, `governor-ratification-backstop` |
| `docs-check.yml` | **`secrets-scan`** (gitleaks) |
| `codeql.yml` | `analyze` |
| `semgrep.yml` | `scan` |
| `benchmark-extraction-gate.yml` | `enforce` |

The governor ratification gate — the mechanism CLAUDE.md describes as the one that stops the governor weakening its own governor — is not a required context. Neither is secrets scanning.

This is #4784's defect class one directory over, on the path where it matters most.

### 2. `require_code_owner_reviews: false`

CODEOWNERS does not gate merges. It auto-requests review; nothing waits for it.

CLAUDE.md states governance changes "require owner ratification and are never auto-merged, **as `CODEOWNERS` already encodes**." Branch protection does not encode that. With `enforce_admins: false` on top, an admin merge bypasses the single required check as well.

I am not claiming anything was subverted — I merged this session's PRs as an approved admin, which is exactly the configured behaviour. The gap is that the *written* guarantee and the *enforced* guarantee differ, and the written one is what a reviewer trusts.

## This corrects my own #4797

#4797 asked you to put `.github/audit-exceptions.json` under CODEOWNERS so security suppressions need owner review. **That would not have worked** — with `require_code_owner_reviews: false`, a CODEOWNERS entry requests a review nobody has to give. I have noted this on that issue; it should be resolved after this one, not before.

## Options (yours to pick)

1. **Add the governor jobs as required contexts** — narrowest fix, closes the governance hole directly.
[...]

PR #4886, opened 2026-08-25
fix(governance): require an approving verdict at the governor review gate

williamzujkowski:
HIGH finding from `.security-discoveries.jsonl`. **Do not merge without owner ratification** — `scripts/check-governor-review.ts` is CODEOWNERS-gated to @williamzujkowski, and it is the governor gate itself. I have not self-merged it.

## The defect

`matchedRecordOutcome` returned `pass` whenever a diff-bound `pr_review` record existed for the PR. It never read `match.verdict`. It did, however, interpolate it into the pass reason — so the gate could literally emit:

```
pass — diff-bound pr_review record found for PR #N (…, verdict=request_changes)
```

**A reviewer who explicitly refused a governor-path change satisfied the gate that exists to require review.** Verdicts are `approve | request_changes | abstain`; all three passed.

No test covered it. The only `request_changes` in the suite uses that value to *break a record hash* for a tamper-evidence case, so every valid-record test is an `approve` and the gap was invisible.

## Decided by vote, because the severity is a real fork

Live 7-voter panel, `higher_order`: **Option A**, 5 of 6 approvers (`request_changes` fails, `abstain` warns).

Nobody argued `request_changes` should pass. The question was whether failing moves ahead of the #4058 enforce flip. The majority reasoning: warn-first is a posture about *review has not happened yet*; a review that happened and refused is the opposite, and the gate already fails closed "regardless of warn-first" for tamper evidence — the other case where it holds affirmative adverse evidence rather than an absence.

The security voter argued Option C (abstain should also fail, since an abstain record is otherwise a bypass). The counter that carried: while absence still only warns, emitting an abstain gains an attacker nothing over emitting no record at all. That bypass becomes real at the #4058 flip, and abstain must harden in the same change — **noted on #4058 rather than left implicit.**

## The dissent found something the majority missed

The contrarian rejected all options over multi-record conflict, and it was right that the case was undefined — though the exposure runs the other way from its guess.

`records.find(...)` returns the **first** match in an append-only ledger, so the *earliest* review for a given diff decided the outcome and every later one was ignored. Two reviewers on the identical diff — one approves, one then refuses — and the gate reported pass.

Verdicts are now aggregated across all matching records with `request_changes` winning, and two tests pin that the result does not depend on ledger order.

## A refusal does not block forever

Worth stating, since a permanently-red gate would be its own defect: records are bound to `reviewedDiffHash`. Pushing a fix changes the diff, the old `request_changes` stops matching, and the gate falls back to warn-on-absence until a new review lands.

## Verification

Six tests, red first, including a positive control that `approve` still passes — failing every verdict would satisfy the others and block all governor-path work.

Three production hunks, each mutation-verified: ignoring the verdict, reverting to first-match-wins, and treating abstain as approval each fail a specific test.

26 tests green. eslint clean. The two `tsc` errors in the repo-root project (`check-harness-alignment.ts`, `website/src/content.config.ts`) are pre-existing on main — confirmed by stashing this branch — and unrelated to these files.

### DIFF
--- a/.changeset/governor-gate-requires-approval.md
+++ b/.changeset/governor-gate-requires-approval.md
@@ -1,0 +1,26 @@
+---
+'nexus-agents': patch
+---
+
+require an approving verdict at the governor review gate
+
+`matchedRecordOutcome` returned `pass` whenever a diff-bound `pr_review` record
+existed for the PR, without ever reading `match.verdict` — and interpolated the
+verdict into the pass reason, so the gate could emit
+`pass ... verdict=request_changes`. A reviewer who explicitly refused a
+governor-path change satisfied the gate that exists to require review.
+
+- `request_changes` now **fails**, closed regardless of warn-first: warn-first
+  covers a review that has not happened yet, not one that happened and said no.
+- `abstain` **warns** — nothing affirmed, nothing refused, so it sits with
+  absence pending the enforce flip (#4058).
+- `approve` passes, unchanged.
+
+Verdicts are also aggregated across every matching record, with
+`request_changes` winning. The gate took `records.find(...)` — the first match
+in an append-only ledger — so the _earliest_ review for a diff decided the
+outcome and an early approve shadowed a later refusal on the identical diff.
+
+A refusal does not block indefinitely: records are bound to
+`reviewedDiffHash`, so pushing a fix changes the hash and the gate falls back
+to warn-on-absence until a new review lands.
--- a/scripts/check-governor-review.test.ts
+++ b/scripts/check-governor-review.test.ts
@@ -225,6 +225,63 @@ describe('analyzeGovernorReview — binding conditions', () => {
     if (outcome.kind === 'fail') expect(outcome.message).toContain('sequence_gap');
   });
 
+  it('(v1) FAILS when the diff-bound record REQUESTED CHANGES', () => {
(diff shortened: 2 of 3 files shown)


williamzujkowski:
The failing check is the Governor-path ratification gate, and that is the correct outcome — not something to fix.

```
- scripts/check-governor-review.ts
These are NEVER auto-merged (CODEOWNERS, CLAUDE.md). To proceed, either
obtain an approving review from a governor-path owner, or have one of them
apply the `owner-ratified` label
```

Worth stating plainly since a red PR usually means "go fix it": every other check passes, and this one is the machinery in #4802/#4690 doing exactly its job on a change to the governor gate itself. **I am not going to admin-merge past it.** The gate says these are never auto-merged; a governor that can route around its own ratification requirement has none.

There is a small recursion worth noticing: this PR strengthens the review gate, and the ratification gate is blocking it until a human approves — which is the same principle applied one layer up. The two are independent checks (ratification is about *who* merges; #4886 is about whether a recorded *verdict* was an approval), so there is no deadlock, just the correct sequence.

@williamzujkowski — this needs either your approving review or the `owner-ratified` label. Note the label only counts when applied by a governor-path owner, per #4690.

williamzujkowski:

Issue #4809, opened 2026-08-25
All 20 Documentation Gate jobs are advisory — the workflow has no aggregate success job


williamzujkowski:
Same defect class as #4784 and #4790, at workflow scale. Distinct from #4802 (which is about branch protection and CODEOWNERS) and separately fixable.

## Finding

`.github/workflows/docs-check.yml` defines **20 jobs**. Verified by parsing both workflows:

```
docs-check jobs:               20
any in ci-success.needs:       []
docs-check has aggregate job:  []
continue-on-error jobs:        ['docs-coverage', 'spell-check']
```

`ci.yml` has a `ci-success` aggregator whose `needs` is the only thing branch protection requires. `docs-check.yml` has **no equivalent**, and none of its jobs are wired into `ci-success`. So every one of them can go red on a mergeable PR:

`skills-index-check`, `agents-index-check`, `repo-index`, `tool-reference-drift`, `strategy-reference-drift`, `canonical-index`, `docs-content-drift`, `registry-coverage`, `typedoc-check`, `schema-fanout`, `secrets-scan`, and nine more.

Two also carry `continue-on-error: true` outright, so they cannot fail even in their own workflow.

## This is not theoretical — it happened today

PR #4808 added a field to the `run` tool schema, which drifted `docs/reference/tools/run.md`. **`Tool Reference Drift` went red and could not have stopped the merge.**

What actually blocked it was `scripts/generate-tool-reference.test.ts`, which asserts the same invariant and runs inside `Script Tests` — a job that became required only yesterday, in #4795. Had that duplicate test not existed, the merge would have landed with the tool reference misdescribing `run`'s interface.

So the drift *was* caught, by luck of duplication rather than by the gate built for it.

## Fix

Mirror the `ci.yml` pattern: add a `docs-success` aggregator to `docs-check.yml` that `needs` its real gates and checks their results, then add **`Docs Success`** to branch protection's required contexts.

Two decisions inside that:

1. **Which of the 20 should be required?** Drift gates over generated artifacts (`tool-reference-drift`, `skills-index-check`, `agents-index-check`, `strategy-reference-drift`, `canonical-index`) are deterministic and cheap — strong candidates. `spell-check` and `docs-coverage` are already marked advisory by their authors and should stay that way, explicitly.
2. **Flake history first**, per the #4784 precedent: do not make a gate required until its behaviour is understood. `secrets-scan` in particular deserves the same care `security` got in #4794.

[...]

PR #4896, opened 2026-08-25
chore(docs): add a Docs Success aggregator and extend the required-job ratchet

williamzujkowski:
Addresses the code half of #4809. CI-only, no shippable source changes.

## Twenty jobs, none of which could block a merge

`docs-check.yml` runs on every PR and defines **20 jobs** — `tool-reference-drift`, `canonical-index`, `governance-drift`, `claims-check`, `secrets-scan` and sixteen more. It had **no aggregator**, and none of its jobs appear in `ci-success.needs`. Branch protection requires named contexts, and there was no single context expressing "the documentation gate passed", so every one of them could go red on a mergeable PR.

**This is not hypothetical.** #4808 added a field to the `run` tool schema, drifting `docs/reference/tools/run.md`. `Tool Reference Drift` went red and could not have stopped the merge. What actually blocked it was `scripts/generate-tool-reference.test.ts` asserting the same invariant inside `Script Tests` — a job that became required only the day before, in #4795. The drift was caught by luck of duplication, not by the gate built for it.

## What this does, and what it deliberately does not

Adds `docs-success` (`Docs Success`), needing all 18 non-advisory jobs and checking each one's `result`.

**It does not make the documentation gate blocking.** Adding `Docs Success` to branch protection's required contexts is owner work, tracked in #4802. Saying that plainly matters, because an aggregator nobody requires is exactly the inert-guard shape this repo keeps finding — I am not going to ship one and imply the hole is closed.

What it does today: collapses twenty scattered results into one honest verdict, and creates the context that protection can require.

## The part that is enforced now

`scripts/ci-required-jobs.test.ts` runs inside `Script Tests`, which **is** required. Extending it to `docs-check.yml` means the aggregator's `needs` list is kept complete by a gate that can actually fail — so the context stays correct in the interval before protection requires it, and a new docs job must declare itself required or advisory rather than defaulting to invisible.

Four assertions, mirroring the `ci.yml` ones: the aggregator exists, every job is classified, every dependency's `result` is checked (a `needs` entry alone only makes the aggregator *wait*), and no advisory job is listed as required.

`docs-coverage` and `spell-check` are excluded deliberately — both carry `continue-on-error: true`, so requiring them would be a gate that cannot fail.

## Verification

Two tests red before the aggregator existed. Both mutations fail the right assertion: dropping a job from `needs` fails the classification test, and removing one `result` check while leaving it in `needs` fails the awaited-but-unchecked test — which is the specific way an aggregator can look complete and tolerate a failure.

The workflow parses (verified with a YAML load, after a first attempt whose generated indentation made `&&` scan as a YAML anchor), and prettier is clean.

## Scope note updated

The file's docstring said it covered `ci.yml` ONLY. It now names both workflows and still says plainly that a green run means "these two are classified", never "every PR gate can block".


williamzujkowski:
The aggregator's first act was to turn this PR red, and that is the outcome I wanted to see.

`Docs Success` failed because `DocOps Skill Sync` failed. Tracing it rather than assuming: `.github/workflows/docs-check.yml` is itself one of the eight tracked pipeline files in `docs/ops/docops-manifest.json`, so changing it requires a corresponding note in `skills/documentation-management/SKILL.md`. The gate is right and the demand is legitimate — I changed the documentation pipeline and did not document it.

Fixed by adding the `PIPELINE NOTE` in the file's established style, and `check-docops-skill.ts` now passes locally.

**Worth stating what this demonstrates, because it is the whole argument for the PR.** `DocOps Skill Sync` lives in `docs-check.yml`, which had no aggregator — so before this change it could go red on a mergeable PR and nobody would be stopped. It has presumably been capable of that for its entire life. The first PR to aggregate the workflow is the first PR the workflow could hold up, and it held up *itself*.

That is a stronger check than the mutation tests: the gate fired on a real omission by the person adding it, not on a synthetic revert.

Still true and unchanged: this does not make the documentation gate blocking. `Docs Success` is not in branch protection's required contexts — that remains #4802 and yours. What is enforced today is that the aggregator stays complete, via `ci-required-jobs.test.ts` inside the required `Script Tests` job.

PR #4903, opened 2026-08-25
fix(doctor): report the fix it made, and a count that matches the verdict

### LINKED ISSUE #4851: doctor reports fixes it did not make and a count that contradicts its own verdict
Two independent defects in `doctor`, both of the "reports success it did not achieve" shape. Filing together because they surface in the same command's output and a reader hits both in one run.

---

## 1. `doctor --fix` claims to have created directories when it created none

`doctor.ts:910-925` runs the data-directory fix when a subdirectory `!exists` **or** `!writable`:

```ts
result.dataDirectory.subdirectories.some((d) => !d.exists || !d.writable)
```

but `initDataDirectories` → `ensureDir` (`setup-data-dir.ts:96-111`) only ever *creates missing* paths:

```ts
if (existsSync(dirPath)) {
  alreadyExisted.push(dirPath);
  return;
}
```

An existing-but-unwritable directory therefore lands in `alreadyExisted`, `failures` stays empty, and `success: true` comes back. Doctor prints `✓ Created missing data directories` and increments `fixCount` **having created nothing and fixed nothing** — for precisely the condition that triggered the fix.

**Observable:** `chmod 500 ~/.nexus-agents/memory`, run `doctor --fix` → `✓ Created missing data directories` and `1 issue(s) fixed`; re-run `doctor` → the identical warning. The unwritable directory is still unwritable.

`initDataDirectories` already returns `created` and `alreadyExisted`; the call site consults neither. The fix is to branch on `created.length` for the message, and to report the unwritable case as unfixed rather than silently counting it as fixed.

## 2. The summary count can contradict the verdict on the same screen

`printDoctorSummary` (`doctor-formatting.ts:384-390`) builds `totalIssues` from CLIs, node version and `mcpServerReady`. `isAllHealthy` (`doctor.ts:801-814`) additionally requires `hasAuthMethod` and `scratchSeverityIsAcceptable(worstSeverity(scratchSpace))`.

So a critical scratch filesystem with everything else healthy takes the **not**-healthy branch and prints:

```
Summary: 0 issue(s) found
```

**Observable:** the count is false, and self-evidently so — a summary printed only because something is wrong, saying nothing is wrong.

Narrower than defect 1: `formatScratchFilesystems` prints the critical filesystem line just above, so the underlying condition *is* disclosed and only the count lies. Still worth fixing, because the count is the line a reader scans.

The two predicates should be derived from one place rather than enumerated twice — that duplication is what let them drift.

---

## Ruled out while verifying

The `versionStatus === 'unsupported'` divergence I first suspected between the same two functions is **not** reachable: `createHealthyResult` (`doctor.ts:371`) sets `authenticated = versionOk && …`, so an unsupported-version CLI is already counted in `unhealthyCount`. Dropped rather than filed.

## Provenance

Both traced against `main`, input and consumer confirmed for each. Same audit series as #4586 and #4671.

williamzujkowski:
Fixes #4851. Two defects in `doctor`, both reporting success it had not achieved.

## 1. `--fix` claimed to create directories when it created none

The fix runs when a data subdirectory is missing **or not writable**:

```ts
result.dataDirectory.subdirectories.some((d) => !d.exists || !d.writable)
```

But `ensureDir` returns early for a path that already exists and never checks writability. So an existing-but-unwritable directory lands in `alreadyExisted`, `failures` stays empty, `success` is true — and doctor printed **"✓ Created missing data directories"** and incremented the fix count, for precisely the condition that triggered it.

`created.length` was already on the result and simply not consulted. A new `describeDataDirFix` reads it and says plainly that setup creates missing directories and cannot repair permissions.

**I nearly shipped an inverted version of this.** My first attempt called `initDataDirectories(true)` as a dry run *after* `runSetup` — by which point setup had created the directories, so the dry run would report zero created and the honest branch would fire on the case it was meant to exclude. It now calls the initializer directly, which is also simpler: `runSetup` was being invoked with every other step skipped anyway.

## 2. The summary count contradicted the verdict

`totalIssues` enumerated CLIs, node version and `mcpServerReady`. `isAllHealthy` additionally fails on an unacceptable scratch severity. So a critical scratch filesystem with everything else healthy printed:

```
Summary: 0 issue(s) found
```

A summary shown *only because something is wrong*, saying nothing is wrong.

The count now includes the same term the verdict reads. `scratchSeverityIsAcceptable` moved from `doctor.ts` to sit beside `worstSeverity` in `doctor-scratch-space.ts`, so the count and the verdict share one definition — the duplication is what let them drift, and two copies would drift again.

## Testing

Five tests, red first, each with its pair: a fix that *is* real must still be counted, and the CLI/node/MCP terms must still be counted rather than replaced by the scratch term.

Three production hunks mutation-verified — dropping the scratch term, counting a no-op as a fix, and counting every outcome each fail a specific test.

**3651 tests across `src/cli`, green.** `tsc`, eslint and the export ratchet clean.

## One thing I did not fix, and it is worth knowing

`#4851` also noted a suspected `versionStatus === 'unsupported'` divergence between the same two functions. It is **not reachable**: `createHealthyResult` sets `authenticated = versionOk && …`, so an unsupported-version CLI is already counted in `unhealthyCount`. Dropped rather than "fixed", so nobody re-derives it later.

### DIFF
--- a/.changeset/doctor-honest-fix-count.md
+++ b/.changeset/doctor-honest-fix-count.md
@@ -1,0 +1,28 @@
+---
+'nexus-agents': patch
+---
+
+stop doctor reporting a fix it did not make and a count that contradicts its verdict
+
+Two defects in `doctor`, both reporting success it had not achieved.
+
+**`--fix` claimed to create directories when it created none.** The fix runs when
+a data subdirectory is missing **or not writable**, but `ensureDir` returns early
+for a path that already exists and never checks writability. An
+existing-but-unwritable directory therefore lands in `alreadyExisted`, `failures`
+stays empty, and `success` is true — so doctor printed
+"✓ Created missing data directories" and incremented the fix count for precisely
+the condition that triggered it. `created.length` was already on the result and
+was not consulted; a new `describeDataDirFix` now reads it and says plainly that
+setup cannot repair permissions.
+
+**The summary count contradicted the verdict.** `totalIssues` enumerated CLIs,
+node version and `mcpServerReady`, while `isAllHealthy` additionally fails on an
+unacceptable scratch severity. A critical scratch filesystem with everything else
+healthy printed `Summary: 0 issue(s) found` — a summary shown only because
+something is wrong, saying nothing is wrong. The count now includes the same
+term the verdict reads, and `scratchSeverityIsAcceptable` moved next to
+`worstSeverity` so both callers share one definition rather than the count
+drifting from the verdict again.
+
+Fixes #4851.
--- a/packages/nexus-agents/src/cli/doctor-formatting.test.ts
+++ b/packages/nexus-agents/src/cli/doctor-formatting.test.ts
@@ -66,6 +66,27 @@ vi.mock('../cli-adapters/types.js', () => ({
   },
 }));
 
+/** Applies a scratchSpace override outside the fixture literal (keeps its complexity under the cap). */
+function withScratch(
+  override: DoctorResult['scratchSpace'] | undefined,
+  base: DoctorResult
+): DoctorResult {
+  return override === undefined ? base : { ...base, scratchSpace: override };
+}
+
+const DEFAULT_SCRATCH_SPACE: DoctorResult['scratchSpace'] = [
+  {
+    label: 'nexus' as const,
+    root: '/tmp/nexus-test',
+    available: true,
+    freeBytes: 20 * 1024 ** 3,
+    totalBytes: 32 * 1024 ** 3,
+    percentUsed: 38,
+    severity: 'ok' as const,
+    message: '20.0 GiB free of 32.0 GiB (38% used)',
+  },
+];
+
 describe('doctor-formatting', () => {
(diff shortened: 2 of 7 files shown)


williamzujkowski:
The Producer/Consumer ratchet was right, and precisely so.

`describeDataDirFix` passed — it has a real cross-file consumer in `doctor.ts`. What failed was the **type**:

```
✗ packages/nexus-agents/src/cli/setup-data-dir.ts :: DataDirFixOutcome
```

`doctor.ts` imports the function, not the interface. The return type is inferred at the call site, so exporting the interface added a public name nothing outside the module names. Made it internal; TypeScript still types the consumer correctly.

Worth noting the gate's precision: it distinguished the exported function (consumed) from the exported type declaring that function's return (not consumed), in the same commit and the same file. A coarser check would have passed both on the strength of the function's import.

Same gate, same lesson as #4847 earlier today — the two escapes it offers are "wire it up" or the `@export-no-consumer-yet` marker, and neither fit. The right answer both times was that the export itself was unnecessary.

8 tests green, `tsc` and eslint clean, ratchet clean locally.

Issue #4919, opened 2026-08-25
p1(governance): Self-Dogfood Review has never run a review and reports success on every PR


williamzujkowski:
`Self-Dogfood Review` has never reviewed a pull request. It reports success on every one, and on at least one run it printed a completed-review verdict over an npm resolution error.

This is the mission statement's stated worst case — *"a record that misreports is worse than a missing one: it launders unreviewed work as reviewed"* — so filing it as a p1 correctness bug on the governor path, not a CI nit.

### 1. The CLI is never runnable, so the review never runs

From run 32872097599 (reviewing PR #4914), verbatim:

```
Routing decision:
npm error could not determine executable to run
Dry run failed

Executing review (if CLI available):
npm error could not determine executable to run
```

`pnpm build` compiles the workspace but never links the binary, so `npx nexus-agents` cannot resolve it. Every run since this workflow landed has failed here.

### 2. The failure is then reported as a completed review

```bash
npx nexus-agents orchestrate "$TASK" --format=json > review-result.json 2>&1 || true

if [ -f review-result.json ] && [ -s review-result.json ]; then
  cat review-result.json
  echo "review_available=true" >> $GITHUB_OUTPUT
else
  echo "review_available=false" >> $GITHUB_OUTPUT
fi
```

`2>&1` redirects the npm error **into** `review-result.json`. The file is therefore non-empty, `-s` passes, and the step reports `review_available=true` — which the summary renders as `Review completed`.

The emptiness check is standing in for a success check. The one condition it cannot distinguish is the one that actually occurs: a file containing an error message. The `No CLI available` branch it was written for is unreachable, because a failing command still writes to the file.

### 3. Even if it ran, it would not read the diff

```bash
PR_DIFF=$(gh pr diff "$PR_NUM" ...)   # assigned, never referenced again
```

`TASK` interpolates `$PR_FILES` — a list of up to 20 **file paths** — and never `$PR_DIFF`. So the reviewer would be asked to assess code quality, security, and test coverage from a list of filenames.

That is the doctrine violation named in `CLAUDE.md`: *"A review must consume the artifact, not a description of it. An agent reviewing a summary of a change has not reviewed the change."* A filename list is not even a summary.

## Why nothing caught it

`continue-on-error: true` on the review step, `|| true` on the command, and `if: always()` on the summary. There is no input that makes this job fail — it is a green check on every PR whose green means only that the runner started.

## What a fix needs

[...]

PR #4920, opened 2026-08-25
fix(ci): stop the self-dogfood review reporting reviews it never ran

williamzujkowski:
Refs #4919. **Owner ratification required** — `.github/workflows/` is `@williamzujkowski` in CODEOWNERS, and this touches a review gate. Not self-merging.

## What was happening

`Self-Dogfood Review` has run on every PR in this repo, including all of mine today, and has never reviewed one. Three defects compounded.

**The CLI was never runnable.** `pnpm build` compiles the workspace but does not link the binary, so `npx nexus-agents` failed with `npm error could not determine executable to run`. Verbatim from run 32872097599, reviewing PR #4914:

```
🔄 Routing decision:
npm error could not determine executable to run
⚠️ Dry run failed

📝 Executing review (if CLI available):
npm error could not determine executable to run
```

**The failure was then reported as a success.** `> review-result.json 2>&1` redirected the npm error *into* the file the script used as its success signal:

```bash
if [ -f review-result.json ] && [ -s review-result.json ]; then
  echo "review_available=true" >> $GITHUB_OUTPUT     # → "✅ Review completed"
```

An emptiness check standing in for a success check. The `No CLI available` branch it was written for is unreachable, because a failing command still writes to the file — so the only condition that ever occurs is the one it cannot distinguish.

**It never read the diff.** `PR_DIFF` was fetched and never referenced; the prompt interpolated `$PR_FILES`, a list of up to 20 paths. The reviewer was asked to judge code quality, security, and test coverage from filenames — the exact thing `CLAUDE.md` rules out: *"A review must consume the artifact, not a description of it."*

## What this changes

**Credentials are checked first, and their absence is reported as unmeasured.** The job passes no model keys today, so with the other bugs fixed it would still fail — just later and more confusingly. `pr-review.yml` already has this guard; this adopts the same shape, reporting rather than failing, since making the gate blocking is a separate policy call.

**The diff is the artifact.** Bounded at 200KB with the covered portion stated in the summary (`diff read: first 200000 of 340112 bytes`) — `CLAUDE.md` permits a bounded read and requires the record to say which part.

**The CLI is invoked through its built entry point**, `node packages/nexus-agents/dist/cli.js`, with an explicit error if the build did not produce it.

**Outcomes are distinguished**: `reviewed`, `no_diff`, `cli_missing`, `cli_failed`, `malformed_output`. Exit status is checked *and* the JSON is parsed and required to be an object — a non-empty file is no longer evidence.

**The default flipped.** With no outcome reported, the summary now says NOT REVIEWED. Previously the absence of a signal rendered as a completed review.

## Verification

I cannot run this end-to-end without a model key, so I tested the two pieces of logic that carry the fix directly:

| input to the JSON gate | verdict |
| --- | --- |
| `npm error could not determine executable to run` | **flagged malformed** (this is the case that used to pass) |
| `{"findings":[]}` | accepted |
| `null` / `123` / `"oops"` / empty | flagged malformed |

And every summary branch, including the two defaults:

```
[reviewed]         → ✅ Reviewed — diff read: all 1200 bytes
[cli_failed]       → ⚠️ NOT REVIEWED — orchestrate failed. See the step log.
[malformed_output] → ⚠️ NOT REVIEWED — orchestrate returned output this job could not parse.
[no outcome, no key] → ⚠️ NOT REVIEWED — no model API key is configured, so no review ran.
[no outcome, key]    → ⚠️ NOT REVIEWED — the review step did not report an outcome.
```

The workflow parses under `yaml.safe_load`, passes `yaml-lint`, and is Prettier-clean.

One deliberate simplification: I dropped an initial `outcome=not_run` write, because it would have depended on GitHub taking the *last* value for a repeated output key. The empty-output default in the summary covers the same ground without that assumption.

## Still open, and deliberately not in this PR

The job remains `continue-on-error: true`. Making a review gate blocking changes merge policy and is yours to decide — I would rather it report honestly first and then be promoted, than be promoted while it still cannot tell a review from an error. #4464 tracks the general advisory→enforcing question.


williamzujkowski:
Extended after a sweep of the other 24 workflows for the same shape. Two additions, both in this file.

**The triage job had the identical defect in a purer form.** `.github/workflows/self-dogfood.yml`, job 2:

```bash
npx nexus-agents orchestrate "$TASK" --dry-run --verbose || true
echo "triage_complete=true" >> $GITHUB_OUTPUT
```

The flag is written **unconditionally**, after `|| true`. `triage_complete` cannot hold any value other than `true`, and the summary prints "✅ Triage analysis complete" off it. Where the review job at least had a condition that was merely wrong, this one has no condition at all.

It is also `--dry-run`, so even on the success path it prints a routing decision and never triages. Same treatment as job 1: credential gate, built entry point, exit status plus JSON validation, and an outcome vocabulary whose default is NOT TRIAGED. The summary now also states plainly that the analysis is advisory and applies no label — which was true before and unstated.

**Both jobs drop to read-only.** `grep` for `gh pr comment` / `gh issue comment` / `gh issue edit` / `--add-label` across the file returns nothing: neither job has ever written anything, yet job 1 held `pull-requests: write` and job 2 held `issues: write`.

That matters more once this PR adds model credentials. A job that processes a PR diff or an issue body (untrusted input), holds repository write access, and carries secrets is the Rule of Two violation `CLAUDE.md` names — all three at once. Dropping to `read` removes the leg that was never used, rather than adding an exception.

williamzujkowski:
## Consensus vote: no consensus — holding for the owner

`higher_order`, supermajority threshold. 6 of 7 approved, but the option split tied **3–3** between "merge both" and "merge #4928, hold this one." Neither cleared the bar, so there is no mandate to merge.

#4928 was merged on the intersection — every approver's option included it. **This PR is the contested half**, and the split is exactly along the conflict-of-interest line.

**For merging** (architect, PM, scope steward): CODEOWNERS enumerates three never-auto-merged governor paths and `self-dogfood.yml` is deliberately not among them; the ongoing harm is concrete — this gate has recorded a successful review on every PR it never reviewed, including all of today's merges — and "strengthening" is mechanically checkable rather than self-asserted (the npm error was redirected into the success file).

**For holding** (security, devex, ai_ml): `self-dogfood.yml` is the gate that reviews agent-authored PRs, and the author is an agent whose PRs it judges. As the ai_ml voter put it:

> "this change strengthens the gate" is the agent's own characterization of its own diff, which is exactly the claim that needs independent eyes.

And on cost asymmetry: the gate has been inert on every PR ever, so a few more days of holding costs nearly nothing, while merging sets a precedent that agents may self-modify their reviewer whenever they argue the change is strengthening.

## Why I am holding rather than breaking the tie

The owner delegated the *merge/hold* decision to the panel, and the panel deadlocked. On a governance-adjacent question with no majority, the conservative branch is the one that preserves the review layer's independence — and I am the interested party, so breaking a tie in favour of merging my own change to my own reviewer is precisely what the dissent warns against.

**@williamzujkowski** — this one wants your call. The panel's own suggested remedy, if you agree with the dissent: add `.github/workflows/self-dogfood.yml` to the governor section of CODEOWNERS, which makes the hard bar explicit rather than a matter of per-PR judgement.

Also worth noting: the panel flagged that the *real* acceptance test for this fix is a live mutation run — a PR the gate should fail, observed to fail. That is best witnessed at ratification time, which is another argument for the hold.

Issue #4927, opened 2026-08-25
Sweep results: eight CI gates whose failure is indistinguishable from success


williamzujkowski:
Sweep of all 25 workflows in `.github/workflows/` for the shape found in #4919: a gate whose failure path is indistinguishable from its success path. Same summary form as #4671 and #4827; each entry below is independently actionable and gets its own issue when picked up.

Every finding was read in the file, and where the verdict lives in a script, in the script.

## Two systemic shapes

**A. `|| echo <default>` makes a fetch failure indistinguishable from a clean measurement.** Findings 1, 5, 7.

**B. A verdict flag set unconditionally, or computed and discarded.** Findings 2, 3, 6, 8.

## Findings

**1. `pricing-drift.yml` — the workflow has no failing path at all.** `DRIFT=$(echo "$OUTPUT" | grep -oP '(\d+) field' … || echo "0")` (L24-28), consumed only by `if: steps.drift.outputs.drift_count > 5`. `scripts/check-pricing-drift.ts` exits **0** on a catalog-fetch failure (L138-139), prints "Non-blocking: script always exits 0" (L183), and exits 0 on any thrown error (L186-189). A litellm outage or a schema rename produces no `N field` text → `drift_count=0` → green weekly run, forever. Sibling `parameter-drift.yml` emits a `skipped` sentinel for exactly this case — that is the shape to copy. Separately, 1-5 genuinely drifted price fields are silently green.

**2. `sica-test-generation.yml` — the only gate on machine-written tests cannot fail.** L165-171, step *"Validate generated tests"*: `pnpm test 2>&1 || echo "Some tests may need adjustment"`. It is the sole check between the generator and `Create Pull Request` (L173), which is conditioned only on `test_count > 0`. Generated tests that fail produce a green "Validate" step and an open PR. L151 swallows the generator's own exit code the same way.

**3. `deploy-website.yml` — `link-check` is in `needs` but cannot fail.** L88 sets `fail: false` on lychee, so its only substantive step never exits non-zero; `deploy` (L91) lists `needs: [build, link-check]` and therefore reads as link-gated while being nothing of the kind. `link-check.yml` uses `fail: true` on the same action — the two disagree about the same check.

[...]

PR #4928, opened 2026-08-25
fix(ci): give the pricing-drift check a verdict a failure can reach

williamzujkowski:
Refs #4927 (finding 1). **`.github/workflows/` is CODEOWNERS-gated — owner ratification required.** Not self-merging.

## The workflow had no failing path

`check-pricing-drift.ts` exits 0 on every route and says so in its own output: `Non-blocking: script always exits 0`. So the exit code carries no signal, and the workflow recovered a verdict by scraping the report prose:

```bash
DRIFT=$(echo "$OUTPUT" | grep -oP '(\d+) field' | grep -oP '\d+' || echo "0")
```

A catalog fetch failure prints **no report**, so there is no `N field` text, so the grep fails, so `|| echo "0"` supplies zero — and the only consumer is `if: drift_count > 5`.

A litellm outage, a schema rename on their side, and genuinely correct pricing all produce the identical green weekly run. That is the reference shape: the failure text lands in the same variable the success signal is read from.

## The sibling already had the answer

`parameter-drift.yml` emits a `PARAM_DRIFT_STATUS` sentinel that **defaults to `skipped`**, and warns loudly when it fires. This is the same treatment applied to pricing:

- The script prints `PRICING_DRIFT_STATUS=clean|drift|skipped` and `PRICING_DRIFT_COUNT=<n>` on **every** terminating path — the fetch-failure branch, the no-drift return, the report path, and the top-level `.catch`.
- The workflow reads those, defaulting the status to `skipped` rather than to a measurement, so an absent verdict is reported as absent.
- The issue gate now requires `drift_status == 'drift'`. Without that term, a skipped run with a defaulted count and a real run with a real count are the same input.

## Two reporting gaps close with it

A skip now emits a `::warning::` instead of passing silently — the point of a sentinel is that someone sees it.

And drift of **1 to 5 fields** now gets a warning. That is real, measured drift which sat below the issue threshold and said nothing at all; the threshold governs whether an issue is filed, not whether the finding is mentioned.

## Tests and mutations

Nine tests: that every terminating path in the script prints a status, that both failure paths print `skipped` specifically (never `clean`), and that the workflow defaults to `skipped`, warns on it, gates the issue on a measured verdict, and reports sub-threshold drift.

| mutation | result |
| --- | --- |
| fetch-failure status line removed | 2 failed |
| catch-handler status line removed | 2 failed |
| workflow default flipped `skipped` → `clean` | 1 failed |
| status term dropped from the issue gate | 1 failed |
| loud-skip step disabled | 2 failed |

## Scope

Seven more findings from the same sweep are in #4927, each independently actionable. I picked this one first because it is the only one where the *entire workflow* has no failing path — the others have a reachable failure somewhere.

### DIFF
--- a/.changeset/pricing-drift-verdict.md
+++ b/.changeset/pricing-drift-verdict.md
@@ -1,0 +1,23 @@
+---
+'nexus-agents': patch
+---
+
+stop a litellm outage reading as clean pricing
+
+CI-only. The pricing-drift workflow had no failing path. `check-pricing-drift.ts`
+exits 0 on every route and says so, so the exit code carried no signal — and
+the workflow recovered its verdict by scraping `"N field"` out of the report
+prose with a `|| echo "0"` fallback. A catalog fetch failure prints no report,
+so it fell through to that fallback and rendered as zero drift. A provider
+outage and correct pricing produced the same green weekly run.
+
+The script now prints `PRICING_DRIFT_STATUS=clean|drift|skipped` and
+`PRICING_DRIFT_COUNT=<n>` on every terminating path, including the top-level
+error handler, and the workflow defaults the status to `skipped` rather than to
+a measurement. `check-parameter-drift.ts` already had exactly this shape; this
+is the same treatment applied to its sibling.
+
+Two reporting gaps close with it: a skip now emits a loud `::warning::` instead
+of passing silently, and drift of one to five fields — real drift that sat
+below the issue threshold and said nothing at all — is now reported without
+filing an issue.
--- a/.github/workflows/pricing-drift.yml
+++ b/.github/workflows/pricing-drift.yml
@@ -24,13 +24,29 @@ jobs:
         run: |
           OUTPUT=$(pnpm check:pricing-drift 2>&1)
           echo "$OUTPUT"
-          # Extract drift count from output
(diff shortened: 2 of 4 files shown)


williamzujkowski:
## Consensus vote: merged on the intersection of both surviving options

`higher_order`, supermajority threshold. Result recorded as **rejected**: 6 of 7 approved, but the option split was a dead tie — **A (merge both) 3, C (merge #4928, hold #4920) 3** — so neither cleared the bar.

That is a no-consensus on the question as asked. But the tally is unambiguous about *this* PR: **A and C both include merging #4928, so all 6 approvers favour it.** The tie is entirely about #4920. Merging here is reading the tally, not overriding it.

The reasoning that separated the two PRs, from the C voters:

> `pricing-drift` is a data-freshness instrument — it measures model-pricing skew, it does not review agent work. No reading of "the governor must not weaken its own governor" reaches it.

And from the A voters, on why the enumeration is the design:

> `check-governor-review.ts` parses the governor section from a literal marker to an explicit `# END governor-owned paths` sentinel (#4683), fail-closed in both directions. #4635 built a parseable single source of truth precisely so "which gates carry the hard bar" is not decided by per-PR intuition.

Neither this PR's files nor `#4920`'s appear in that section. The broad `/.github/workflows/` rule is review-requested, not merge-blocked.

**#4920 is held** — see its own comment.

PR #4930, opened 2026-08-25
fix(config): recognize four documented env vars, and isolate the suite's data dir

### LINKED ISSUE #4722: test-hygiene: the suite writes to the developer's real homedir data dir
Found while adding the tool-fitness producer (#4656). **Pre-existing and broader than that change**, so filing separately rather than fixing inline.

## What happens

Running `npx vitest run` writes to `~/.nexus-agents/` — the real, homedir-scoped, cross-repo data dir. Observed during one full-suite run today:

```
~/.nexus-agents/capability-gaps.jsonl   modified 06:00
~/.nexus-agents/memory                  modified
~/.nexus-agents/learning                modified
~/.nexus-agents/tool-fitness            created
```

Only the last is new (from #4656). The first three predate it, so this is an existing pattern that the new producer joins rather than introduces.

## Why it matters

This directory holds **cross-repo governance and learning data** — capability gaps, memory, outcome/learning state. It is deliberately homedir-scoped so signal accumulates across repositories.

Consequences:

1. **Test runs contaminate real signal.** Synthetic tool names, fabricated capability gaps, and test outcomes land in the same store the routing and improvement loops read. A `demo_tool` or `throwing_tool` entry is indistinguishable from a real one to a consumer.
2. **It is silent.** Nothing warns, and the data is outside the repo, so `git status` never shows it.
3. **CI runners accumulate it too**, though there it is discarded with the runner.

The tool-fitness ledger's own module doc already anticipates this — *"Tests that DO exercise persistence should construct their own `ToolFitnessLedger` with an explicit `filePath` to stay isolated"* — but that is per-module guidance, honoured unevenly, and it cannot cover code that reaches a singleton indirectly through middleware.

## Suggested fix

A global test setup pinning `NEXUS_DATA_DIR` to a per-run temp directory, alongside the existing `cli-spawn-guard.setup.ts` in `vitest.config.ts` `setupFiles`. That makes isolation the default rather than each module's responsibility, which is the same reasoning that made the CLI spawn guard a setup file rather than a convention.

Two things to check before doing it:

- Some tests may depend on a stable data dir across the run; a per-run (not per-file) temp dir preserves that.
- Some may read fixtures from the real dir and would need explicit paths.

Worth measuring how many tests change behaviour under an isolated dir before committing to the approach — if the answer is "many", that is itself a finding about hidden coupling to developer-local state.

## Interim

#4656's own tests use a throwaway `NEXUS_DATA_DIR`, and the pollution created while developing it was removed. That covers one module, not the suite.

williamzujkowski:
Closes #4722.

## The reported problem

`npx vitest run` wrote to `~/.nexus-agents/` — the real, homedir-scoped, cross-repo store holding capability gaps, memory and learning outcomes. Test runs put synthetic tool names and fabricated gaps into the same data the routing and improvement loops read. Silently, and outside the repo, so `git status` never showed it.

The fix is one line in `vitest.config.ts`, beside the `TMPDIR` redirect that was already there. Config-level for the same reason the CLI spawn guard is a setup file and not a convention: code reaching a singleton through middleware cannot opt in. Per-run rather than a fixed path, since a shared directory is the same state-leaking hazard as the homedir, only narrower; it sits under the scratch root so the existing reaper collects it.

## The issue asked me to measure first, and the measurement was the interesting part

Its own suggested first step was *"measure how many tests change behaviour under an isolated dir before committing to the approach — if the answer is many, that is itself a finding about hidden coupling to developer-local state."*

**Thirteen tests across four files.** All the same shape: they isolate by mocking `homedir()`, which only works while the data dir is homedir-derived. `NEXUS_DATA_DIR` outranks that branch.

Which means — and this is the part that outlives the test-hygiene fix — **those tests already fail on `main` today for anyone who sets `NEXUS_DATA_DIR`.** Verified directly, no changes of mine:

```
main, no env:                    63 passed
main, NEXUS_DATA_DIR=/tmp/probe: 13 failed | 43 passed
```

A documented, first-class environment variable breaks the suite. They now isolate via `NEXUS_DATA_DIR` itself, so they no longer depend on which branch the resolver takes.

## And the variable was not in the schema at all

The full-suite run surfaced `env-schema.test.ts` failing with `expected [ { name: 'NEXUS_DATA_DIR' } ] to have length 0`. `validateNexusEnv` classified it as an **unknown** `NEXUS_*` var — a typo suggestion, for a name spelled correctly, taken from CLAUDE.md's own table.

Checking the rest of that table found four missing: `NEXUS_DATA_DIR`, `NEXUS_REPO_PREFERRED`, `NEXUS_ACCESS_POLICY_MODE`, `NEXUS_SANDBOX`. Its sibling `NEXUS_TMPDIR` was there, which is what makes the omission look accidental rather than deliberate.

## Three ratchets, because each fix is one deletable line

| test | catches |
| --- | --- |
| documented vars ⊆ schema | the docs and the schema drifting apart again |
| `NEXUS_DATA_DIR` is set for every worker | the config line being dropped |
| …and does not resolve under `~/.nexus-agents` | a value that satisfies the above and isolates nothing |

The first excludes variables CLAUDE.md documents *as removed* (#2977, #4180) and carries a guard-the-guard assertion, since a scan matching nothing would pass over an empty set — the exact failure this file is about.

## Verification

| | |
| --- | --- |
| full suite | **28,475 passed**, 1 failed → 0 after the last fix |
| writes to `~/.nexus-agents/` during the run | **none** (`find -newer` over the window) |
| drop `NEXUS_DATA_DIR` from the config | isolation test fails |
| drop `NEXUS_DATA_DIR` from the schema | 8 tests fail |
| make the docs scan match nothing | guard-the-guard fails |

One process note: I lost the ratchet test once to a `git checkout -- <file>` used to undo a mutation, which reverted the real work alongside it. Restored from the same source; recording it because it is a trap I have hit before.

### DIFF
--- a/.changeset/test-data-dir-isolation.md
+++ b/.changeset/test-data-dir-isolation.md
@@ -1,0 +1,23 @@
+---
+'nexus-agents': patch
+---
+
+recognize NEXUS_DATA_DIR, NEXUS_REPO_PREFERRED, NEXUS_ACCESS_POLICY_MODE and NEXUS_SANDBOX in the env schema
+
+All four are documented in CLAUDE.md's environment table and none were in
+`env-schema.ts`, so `validateNexusEnv` reported them as **unknown** `NEXUS_*`
+variables — offering a typo suggestion for a name spelled correctly. Anyone
+following the documentation to set a runtime data root was told the variable
+did not exist.
+
+Found while stopping the test suite writing to `~/.nexus-agents/`, the real
+cross-repo store holding capability gaps, memory and learning outcomes: test
+runs were putting synthetic tool names and fabricated gaps into the data the
+routing and improvement loops read. The suite now runs against a per-run data
+dir set in `vitest.config.ts`, beside the `TMPDIR` redirect that was already
+there.
+
+A test now cross-checks the documented list against the schema, so the two
+cannot drift apart again, and another asserts the suite's data dir is not the
+real one — the isolation is a single config line, and nothing else would notice
+it being dropped.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -72,6 +72,8 @@ Most-used env vars:
 
 Full list in [docs/getting-started/CONFIGURATION.md](./docs/getting-started/CONFIGURATION.md). Install: [INSTALLATION.md](./docs/getting-started/INSTALLATION.md). Sandboxed: [SANDBOXED-USAGE.md](./docs/guides/SANDBOXED-USAGE.md).
 
+Every `NEXUS_*` variable is validated at startup against `config/env-schema.ts`; an unrecognized name is reported as unknown with a typo suggestion. A variable named in this table but missing from that schema is therefore reported as a typo despite being spelled correctly (#4722), so the two lists are cross-checked by a test.
+
 Note: `NEXUS_WORKERS_*` / `NEXUS_WORKFLOW_MAX_PARALLEL` / `NEXUS_TEST_PARALLELISM` / `NEXUS_EVALUATION_MAX_WORKERS` / `NEXUS_EVENTBUS_MAX_HISTORY` / `NEXUS_SWARM_OBSERVER_MAX_EVENTS` were removed in 2.82.0 (#2977 — silent no-ops; consumer wiring never landed); `NEXUS_TEST_TIMEOUT_MS` / `NEXUS_TIMEOUT_CLISIMPLE` / `NEXUS_TIMEOUT_CLICOMPLEX` were removed for the same reason in #4180 (per-complexity CLI timeouts flow through `TIMEOUT_PROFILES`, not env vars).
 
 ---
--- a/docs/getting-started/CONFIGURATION.md
+++ b/docs/getting-started/CONFIGURATION.md
@@ -168,6 +168,14 @@ logging:
 
 All configuration can be overridden with environment variables:
 
+**Every `NEXUS_*` variable is validated at startup** against
+`packages/nexus-agents/src/config/env-schema.ts`. A name the schema does not
+recognize is reported as unknown, with a typo suggestion — including a name
+that is correct and documented here but absent from the schema, which is how
+`NEXUS_DATA_DIR` came to be flagged (#4722). A variable added to this document
+must be added to the schema in the same change, and a test now cross-checks the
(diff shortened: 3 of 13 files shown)


williamzujkowski:
CI turned up something worth flagging before this merges.

**`Registry-Coverage Check` failed, and its own rationale names this exact variable.** From the manifest:

> Receipts: #2315 — env-var rollouts (**NEXUS_DATA_DIR**) miss callsites and docs.

The gate requires that a change to `env-schema.ts` also touch `CLAUDE.md` and `docs/getting-started/CONFIGURATION.md`. But all five variables involved were already in both:

```
NEXUS_DATA_DIR               CLAUDE.md=yes CONFIG.md=yes
NEXUS_REPO_PREFERRED         CLAUDE.md=yes CONFIG.md=yes
NEXUS_ACCESS_POLICY_MODE     CLAUDE.md=yes CONFIG.md=yes
NEXUS_SANDBOX                CLAUDE.md=yes CONFIG.md=yes
NEXUS_SANDBOX_ROOT           CLAUDE.md=yes CONFIG.md=yes
```

The drift ran the other way — documented and missing from the schema — so the gate fires on a PR that *closes* drift and has no documentation change to make. Filed as #4931; I did not reach for the manifest escape hatch, because weakening a gate to land my own PR is the wrong instinct to build in.

**What I added instead is a real gap, not a cosmetic touch.** Nothing in either document said the variables are validated at all — which is precisely why a correctly-spelled, documented variable being reported as a typo is so confusing. Both peers now say so, and name the failure mode.

**Two things for you specifically:**

`CLAUDE.md` is a required peer and is `@williamzujkowski` in `CODEOWNERS`, so **this coupling forces every `env-schema.ts` change through owner ratification** — including adding a variable the docs already describe. That may be intended, since env declarations are arguably governance surface. If it is not, `CONFIGURATION.md` is the full reference and `CLAUDE.md` carries a curated most-used subset that will never list every variable, which makes it the wrong required peer. #4931 lays out three options.

The `CLAUDE.md` edit is one sentence, outside the `inject-governance.ts` generated blocks. `pnpm governance:check` passes (47 tools / 12 experts / 33 skills unchanged) and markdownlint is clean.

williamzujkowski:

Issue #4910, opened 2026-08-25
fix(routing): LinUCB trains on a different feature vector than it scores with


williamzujkowski:
Found by an adversarial review of my own same-day merges, and it is a regression I introduced in #4874.

## The invariant

LinUCB is only consistent if the feature vector handed to `update` is the one `selectArm` scored. `contextToFeatures` (`linucb-math.ts:41-49`) puts `budgetUtilization` at index 4 of a six-element vector, and the ridge regression fits `θ` against whatever `update` receives.

## What broke

#4874 gave the **select** path a real value — `composite-router-stages.ts:1114-1120` now passes `budgetResult.value.budgetUtilization` into `runLinUCBStage`.

The **update** path was left as it was. `composite-router-outcome.ts:57`:

```ts
const context = taskAnalysisResultToBanditContext(analysis);
```

No options, so `budgetUtilization` falls to the `NEUTRAL_FEATURE` default of `0.5` (`task-profile-adapter.ts:131`).

Before #4874 both sides were `0.5` and the dimension was inert — dead weight, which is what that issue set out to fix. After it, the column is constant during training and varying at score time. `θ_budget` gets fit against a constant and then multiplied by a value between 0 and 1, arm by arm, with the bias differing by observation count. The dimension went from carrying no information to carrying misinformation.

`seedPriors` (`linucb-bandit.ts:275`) and `warmStart` (`:314`) replay at `0.5` for the same reason.

## Reachability

Narrower than it first looks, and worth stating precisely. `applyBudgetFilter` emits a utilization only when `maxCostUsd` is set (`composite-router-helpers.ts:224-227`), and `nexus-agents.yaml` ships `budget:` commented out. On a default install both paths are still `0.5` and symmetric, so nothing is wrong today unless someone configures a cost ceiling — `--max-cost-usd`, or uncommenting the YAML block.

That also means #4874's own tests could not have caught it: both hand-feed `0.87` to the select path, and nothing exercised the pair.

## Fix

`checkBudget` is a pure function of the task and the constraint, so the outcome path can **reproduce** the selection-time value rather than caching a context across the request boundary. Extract the formula into one place both paths call, thread `budgetRouter` and `budgetConstraints` into `OutcomeDependencies`, and recompute in `recordBanditOutcome`.

Sharing the formula is the point: two call sites computing the same feature independently is how this drifts again.

## The wider lesson

[...]

PR #4944, opened 2026-08-25
feat(audit): carry read coverage on the chain verification verdict

### LINKED ISSUE #4805: verify_audit_chain reports ok:true on a partially-read log; the coverage signal is on a sibling field
Follow-up to #4788, which fixed half of this.

## Finding

#4788 made the tool report `skippedLines` / `unreadableFiles` so a partial read is visible. But the **verdict object itself is unchanged**. My own test asserts the state precisely:

```ts
expect(body.eventCount).toBe(3);
expect(body.verification.ok).toBe(true);   // with skippedLines: 2
```

So a caller reading `ChainVerification` — the object #4768/#4773 hardened *specifically* so `ok: true` could not be mistaken for "verified" — receives a clean, complete-looking verdict over a log two lines of which were never read. The coverage signal sits on a sibling field the caller has to know to check.

`notVerified` was added for exactly this shape of problem. It just does not cover this case.

## Why it's not a one-line fix

`verifyChain(events)` takes only the events. It cannot know lines were dropped — the *loader* knows. And `notVerified` currently means "nothing was verified" (`'empty'` | `'unchained'`), which a partial read is not: real links were checked, just not all of them.

So the options differ in where the honesty lives:

1. **Extend `ChainVerification`** with a coverage field (e.g. `partial: { skipped: number }`) and have the tool populate it. Puts the signal where callers already look. Touches `src/audit/` — governor path, needs ratification.
2. **Have the loader pass a coverage hint into `verifyChain`** so the verdict is computed with knowledge of what was dropped. Cleaner semantically, larger change.
3. **Leave the verdict alone and document the contract**: `verification` answers "do the events I was given chain correctly", and coverage is a separate axis the caller must read. Cheapest, but it is the option that keeps a trap for the caller who reads only `ok`.

I lean 1 — the whole point of the #4773 work was that `ok: true` alone must not read as assurance — but it is a governor-path change and deserves a decision rather than my picking it unilaterally.

## Not urgent, but not cosmetic

No current caller is misled in production that I can find; `verify_audit_chain` is invoked ad hoc. It matters because the audit chain is the instrument a human spot-check trusts most, and per CLAUDE.md a partial review recorded as complete is the failure mode that turns the audit trail into a rubber stamp.

## Provenance

Found by an adversarial review of my own #4788, then confirmed against the test I wrote, which asserts the misleading state as intended behaviour.

williamzujkowski:
Closes #4805. **`src/audit/` is on the never-auto-merge list — owner ratification required.** Not self-merging.

## The decision was the panel's, not mine

The issue offered three options and I deliberately did not pick one. A 7-voter `higher_order` panel with the options declared:

```
decision: approved   approval: 83.3%   5 approve / 1 reject / 0 abstain / 1 error

A: extend ChainVerification with a coverage field the tool populates   4
B: pass a loader coverage hint into verifyChain                        1

optionCoverage: { approverCount: 5, selectedCount: 5, unattributedApprovals: 0 }
```

Every approver named an option, so the 4-1 is a real preference rather than an artifact of the approve/reject bar. Vote record sequence 68.

Worth reading the architect's reasoning before reviewing: **B** was rejected on layering — threading loader state into `verifyChain` mixes the I/O axis (what was read) with the logic axis (do the links hold), and forces the question of what verdict a *valid but partial* chain gets, which mutates existing fields for no gain over A. **C** was rejected because `ChainVerification` is the evidence artifact: it gets serialized, persisted, and passed around **without** its sibling fields, which is exactly when "the coverage is on a sibling" stops being true.

The dissenting voter was `catfish` at 0.95 confidence — a confident contrarian on a governor-path change is the voice the panel exists to surface, and worth your reading directly.

## Three boundaries the implementation holds

**Absent means unknown, not complete.** `verifyChain` receives only events and cannot know what the loader dropped, so it attaches nothing. Defaulting to zero would manufacture precisely the assurance this field exists to stop manufacturing. `skipped: 0` is therefore a *positive claim* of full coverage, which only a caller who read the log can make.

**`notVerified` is not overloaded.** It means nothing was verified (`'empty'`, `'unchained'`). A partial read verified real links — a different axis — so it gets its own field rather than a third enum value that would change what #4768/#4660 gave that one.

**A failing verdict is returned untouched.** It already names a specific event index, and coverage does not qualify a detected break.

## Mutations

| mutation | result |
| --- | --- |
| never attach coverage | 2 failed |
| attach to failing verdicts too | 1 failed |
| tool hardcodes `skipped: 0` | 1 failed |
| tool stops attaching entirely | 2 failed |

The third row needed a second pass. With only the unit tests, hardcoding `skipped: 0` in the tool left **all 318 tests green** — the type-level behaviour was pinned and the wiring was not, the same seam gap as #4910 and #4940. There is now a tool-level test that writes a log with unparseable lines and asserts the verdict itself reports `skipped: 2`, plus its pair asserting a clean log reports `skipped: 0` rather than nothing.

Public API surface unchanged; 320 audit and tool tests pass.

### DIFF
--- a/.changeset/chain-verification-coverage.md
+++ b/.changeset/chain-verification-coverage.md
@@ -1,0 +1,26 @@
+---
+'nexus-agents': minor
+---
+
+carry read coverage on the audit-chain verdict itself
+
+`verify_audit_chain` reported `skippedLines` and `unreadableFiles` as sibling
+fields of its response, but `ChainVerification` — the object a caller reads to
+decide whether tamper-evidence holds — was unchanged. A log two lines of which
+were never parsed still produced a clean, complete-looking verdict.
+
+`ChainVerification` now carries an optional `coverage: { skipped,
+unreadableFiles }`, attached by the tool, which is the only party that knows
+what the loader dropped. `withCoverage` is how a caller states it.
+
+Three deliberate boundaries. Absent coverage means UNKNOWN, not complete —
+`verifyChain` receives only events and cannot know, so defaulting to zero would
+manufacture the assurance the field exists to stop manufacturing. `skipped: 0`
+is therefore a positive claim of full coverage rather than an omission.
+`notVerified` keeps its meaning of "nothing was verified"; a partial read
+checked real links and is a different axis, so it is not overloaded. And a
+failing verdict is returned untouched, since it already names a specific event
+index that coverage does not qualify.
+
+Resolves the #4805 fork by a 7-voter `higher_order` panel: Option A, 4-1 among
+approvers, `unattributedApprovals: 0`.
--- a/packages/nexus-agents/src/audit/audit-chain-verify.test.ts
+++ b/packages/nexus-agents/src/audit/audit-chain-verify.test.ts
@@ -8,7 +8,7 @@
 
 import { describe, it, expect } from 'vitest';
 import * as crypto from 'node:crypto';
-import { verifyChain } from './audit-logger.js';
+import { verifyChain, withCoverage } from './audit-logger.js';
 import type { AuditEvent } from './audit-types.js';
 
 function realHash(event: AuditEvent): string {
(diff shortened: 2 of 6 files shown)


williamzujkowski:

PR #4950, opened 2026-08-25
fix(cli): pad boxes by display width so coloured lines keep their border

### LINKED ISSUE #4913: fix(cli): the routing-audit box pads on ANSI-inclusive length, so five lines render outside the box
Surfaced by a regression test written for a different bug (#4891 pushed an 82-character line into a 63-character box). Widening that test from one stage to the whole report fails on five more lines with a different root cause, so it is filed here rather than folded into the regression fix.

## Root cause

`boxLine` pads on `String.prototype.length` (`box-drawing.ts:38`):

```ts
return color('|', borderColor) + content.padEnd(BOX_WIDTH - 2) + color('|', borderColor);
```

`content` routinely carries ANSI escapes, and `.length` counts them — an escape like `ESC[1m` is five characters that occupy zero columns. So the pad target is wrong by however many escape bytes the caller happened to include, and the right border lands in a different column on every line depending on how many colours it used.

## The tell is already in the source

Three call sites have hand-tuned the fudge factor rather than fixing the padding:

| line | pad target |
| --- | --- |
| `routing-audit-format.ts:121` | `BOX_WIDTH + 8` |
| `routing-audit-format.ts:133` | `BOX_WIDTH + 11` |
| `routing-audit-format.ts:162` | `BOX_WIDTH + 7` |

Each constant is the escape-byte count for that specific line's specific colours. They are correct only until someone adds a colour, changes a marker, or renders with `NO_COLOR` — under which the escapes vanish and every one of these over-pads by 7 to 11 columns.

## Observable failure

`formatAsciiOutput(result, { explain: true })`, visible length after stripping ANSI, against `BOX_WIDTH = 65` — five lines exceed it:

- `Type: code_implementation | Complexity: 6/10 | Tokens: ~5000 | Flags: code` (78 visible)
- the three `UCB: ... pulls: ...` arm rows
- `Final Selection: claude`

Every one is a ragged right border in `nexus-agents routing-audit`.

Two distinct problems are mixed in there, and a fix needs both:

1. **Mis-padding** (the UCB rows, Final Selection) — content fits, the pad target is wrong.
2. **Genuine overflow** (the `Type:` line at 78 visible characters) — content does not fit a 63-column area at all, so no padding scheme saves it. The caller has to split or elide.

## Fix sketch

Give `box-drawing.ts` a `visibleWidth(s)` helper that strips the ANSI SGR pattern before measuring, pad against that, and delete the three `BOX_WIDTH + n` fudge factors. Then handle the one genuinely-too-long line at its source.

Worth adding the general form of the guard once fixed — the scoped version now lives at `routing-audit-format.test.ts` under `keeps every rendered line inside the box`, and widening it from `formatBudgetFilter` to `formatAsciiOutput` is a one-line change that turns it into a permanent ratchet for every `boxLine` caller.

## Why this was invisible

The existing assertions are all `toContain('...')`, which passes on a broken render — they check that text is present, never that the box is intact. Same shape as the rest of the cannot-report-a-negative series (#4671).

williamzujkowski:
Closes #4913, found by a regression test written for a different bug.

## Two problems wearing one symptom

**Mis-padding.** `boxLine` padded with `String.prototype.length`, which counts ANSI escape bytes. An escape occupies zero columns, so every coloured line was padded short by however many it contained. Three call sites had already worked around this by hand:

| line | pad target |
| --- | --- |
| `routing-audit-format.ts:119` | `BOX_WIDTH + 8` |
| `:131` | `BOX_WIDTH + 11` |
| `:160`, `:206`, `:213` | `BOX_WIDTH + 7`, `+ 7`, `+ 5` |

Each constant is the escape-byte count for that line's specific colours — correct until someone adds a colour, and over-padding by 7 to 11 columns under `NO_COLOR`, where the escapes vanish. `visibleWidth` strips SGR escapes before measuring; all five constants are gone and those lines now go through `boxLine` like everything else.

**Genuine overflow.** The task-analysis summary is one pipe-joined string that reached **78 display columns** in a 63-column area. No padding scheme fixes content that does not fit, so `formatTaskAnalysis` wraps it on the `|` separators. A single segment wider than the box is still emitted whole — it overflows visibly rather than being silently truncated, because hiding data is worse than a ragged line.

## The tests needed a second pass, for a reason worth stating

My first mutation run reported the ANSI fix as **unpinned**: reverting `visibleWidth` to `s.length` left all 44 tests green.

That is because `colors.*` are empty strings under `NO_COLOR` and non-TTY — which is exactly how the suite runs. With no escapes present, ANSI-aware and byte-count padding are indistinguishable, so no rendered-output test can tell them apart. The fix is invisible to the environment that tests it.

So `box-drawing.test.ts` now injects the escape sequences literally and asserts a coloured line pads to the same display width as a plain one. Both mutations fail against it:

| mutation | before | after |
| --- | --- | --- |
| `visibleWidth` counts escapes | green | **1 failed** |
| `boxLine` pads on raw `.length` | green | **1 failed** |
| remove the task-summary wrapping | 1 failed | 1 failed |

## Scope

The regression test that surfaced all of this is widened from one section to `formatAsciiOutput` in full, so every `boxLine` caller is covered — that widening was the point of filing #4913 separately rather than folding it into #4914.

6832 CLI tests pass; public API surface unchanged.

### DIFF
--- a/.changeset/box-ansi-aware-padding.md
+++ b/.changeset/box-ansi-aware-padding.md
@@ -22,3 +22,11 @@ emitted whole and still overflows — visibly, rather than silently truncated.
 
 The regression test that found these now covers the whole rendered report
 rather than one section.
+
+`centerText` had the identical defect and is fixed alongside: it computed
+`BOX_WIDTH - text.length - 2`, so coloured text drifted left by half the escape
+length.
+
+`visibleWidth` lives in its own module. Measuring display columns is not
+box-specific, and the producer/consumer ratchet is right that a test importing
+a symbol is not a consumer of it.
--- a/.changeset/box-ansi-aware-padding.md
+++ b/.changeset/box-ansi-aware-padding.md
@@ -1,0 +1,24 @@
+---
+'nexus-agents': patch
+---
+
+measure box padding in display columns, not bytes
+
+`boxLine` padded with `String.prototype.length`, which counts ANSI escape
+bytes — an escape occupies zero columns — so every coloured line was padded
+short by however many escapes it happened to contain and its right border
+landed in the wrong place. Three call sites in the routing audit had grown
+hand-tuned `BOX_WIDTH + 8` / `+ 11` / `+ 7` compensations, each correct only
+for that line's exact colours and all of them over-padding under `NO_COLOR`.
+
+`visibleWidth` strips SGR escapes before measuring, and the five fudge factors
+are gone.
+
+One line overflowed for a different reason: the task-analysis summary is a
+single pipe-joined string that reached 78 display columns in a 63-column box.
+No padding scheme fixes content that does not fit, so `formatTaskAnalysis`
+wraps it on the `|` separators. A single segment wider than the box is still
+emitted whole and still overflows — visibly, rather than silently truncated.
+
+The regression test that found these now covers the whole rendered report
+rather than one section.
--- a/packages/nexus-agents/src/cli/ansi-width.ts
+++ b/packages/nexus-agents/src/cli/ansi-width.ts
@@ -1,0 +1,27 @@
+/**
+ * Display-width measurement for terminal strings.
+ *
+ * Its own module rather than a box-drawing detail: measuring how many columns a
+ * string occupies is not specific to boxes, and `box-drawing.ts` is only its
+ * first caller. Splitting it also gives the export a real cross-file consumer,
+ * which the producer/consumer ratchet is right to demand — a test importing a
+ * symbol is not a consumer of it.
+ *
+ * @module cli/ansi-width
+ */
+
+/** ANSI SGR escapes — zero display columns, non-zero `String.length`. */
+const ANSI_SGR = /\u001b\[[0-9;]*m/g;
+
+/**
+ * Display width of a string, ignoring ANSI colour escapes (#4913).
+ *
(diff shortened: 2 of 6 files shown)


williamzujkowski:
Two additions after the producer/consumer ratchet flagged the first version.

**`visibleWidth` moved to its own module.** It was exported from `box-drawing.ts`, used only by `boxLine` in the same file, and imported by the test — which the ratchet correctly rejects, since a test importing a symbol is not a consumer of it. `ansi-width.ts` is also the better home: measuring how many columns a string occupies is not box-specific, and box drawing is just its first caller.

That is the third time today this ratchet has caught a same-file-only export of mine (after `auditLineFor` and `partitionWarmStartSkips`). It is earning its keep.

**`centerText` had the identical defect**, three lines below the one I was fixing:

```ts
const padding = Math.max(0, BOX_WIDTH - text.length - 2);
```

Same `.length`, same escape-byte miscount — coloured text drifts left by half the escape length. Fixed the same way.

It needs its own assertion: mutating `centerText` back to raw `.length` passed all 48 tests, because everything else in the file exercises `boxLine`. Sibling functions with the same bug do not get covered by fixing one of them — the same lesson as the QA-iteration cap in #4945 earlier today.

| mutation | result |
| --- | --- |
| `centerText` back to raw `.length` | **1 failed** (was: 48 passed) |

6833 CLI tests pass.

PR #4951, opened 2026-08-25
feat(doctor): report when the global install has drifted from this build

### LINKED ISSUE #4767: ops: nothing detects when the globally-installed MCP server falls behind the repo
## The MCP server was six minor versions behind all session

Checked before running an end-to-end validation, per the discipline of not trusting live MCP calls as evidence without confirming what they execute:

```
repo:           3.15.0
npm:            3.15.0
global install: 3.9.4     <- what the MCP server actually ran
```

`.mcp.json` runs off the global install, so every `consensus_vote`, and any other MCP tool called today, executed **3.9.4** — not the working tree, and not anything merged today. Now synced to 3.15.0.

## What this does and does not affect

**Does not affect the votes.** The panels I ran are *decisions*, and their value is the reasoning of seven voters on a written proposal. The version of the tallying code does not change what the architect seat argued. Every load-bearing claim I put in those proposals was verified by reading the working tree directly, not by MCP output.

**Does affect one claim I made.** On #4701 I wrote that a specific panel "was tallied by simple majority despite the `higher_order` label", citing `appliedToDecision: false`. That analysis was correct **for the repo source**, which is what matters for the fix — but I attributed it to the run I had just executed, and that run was 3.9.4. I did not inspect the returned `higherOrderMetadata` to confirm the field was even present in that version. The code analysis stands; the specific attribution to that run was not verified.

**Does affect any future e2e validation.** Running one before this sync would have exercised 3.9.4 and proved nothing about today's twenty merged PRs — which is precisely the trap of treating a live call as evidence.

## Why it drifted

Nothing keeps them in step. The release workflow publishes to npm; the global install is a separate manual `npm install -g`. Six minors accumulated without a signal. Worth a check in `nexus-agents doctor` — filed separately if there is appetite.

williamzujkowski:
Closes #4767. Built because it bit me twice in one day: the global install was **eleven minor versions** behind this morning (4.3.1 vs 4.14.1) and had drifted **three more** by the afternoon.

## Why it matters more than a stale binary

`.mcp.json` runs the MCP server off the global install. So when it drifts, every `consensus_vote`, every research call, every pipeline run executes code the operator is not looking at — and there is no signal at all. The morning's drift meant a full day of MCP calls could not be cited as evidence about the day's merges.

Nothing keeps the two in step: the release workflow publishes to npm, and the global install is a separate manual `npm install -g`.

## The two design decisions

**`unknown` is a state, not a pass.** "No global install found" and "the versions match" are different facts. A check that renders the first as the second reproduces the exact condition that let the drift accumulate — nobody knew, so nobody looked. `installFreshnessIsHealthy` returns true only for `aligned`; mutating it to `state !== 'behind'` fails a test whose comment says why.

**The remedy says restart, not just update.** An already-spawned MCP server keeps the old code until it is restarted, so `npm install -g` alone reports the problem resolved while the session continues on the stale build. I know because that is what I did this morning.

## A test that could not fail, caught by mutating it

The remedy assertion was originally:

```ts
expect(line).toContain(INSTALL_FRESHNESS_REMEDY);
```

Tautological — it compares the rendered line against the same constant it is rendered from, so deleting the restart clause mutates both sides and the test still passes. Verified: that mutation left all 11 tests green. There was a `toMatch(/restart/i)` beside it, which also survived, because the constant's *second* sentence still contained "restarted".

Now asserted against literals (`'npm install -g'`, `'RESTART'`, and the explanatory clause) with no reference to the constant. Deleting the restart guidance fails.

That is the "test supplies the answer" shape, and I would not have found it by reading the test.

## Mutations

| mutation | result |
| --- | --- |
| `unknown` counts as healthy | 1 failed |
| empty version string falls through to `behind` | 1 failed |
| remedy drops the restart clause | 1 failed *(after the fix above; passed before)* |

## Shape

The verdict is a pure function in `doctor-install-freshness.ts`; `doctor.ts` holds only the `npm ls -g` call and injects it, so the honesty logic is testable without npm. The reader tolerates unparseable output — `npm ls -g` prints warnings to stdout in some configurations, and a crash there would take `doctor` down.

6843 CLI tests pass; public API surface unchanged.

### DIFF
--- a/.changeset/doctor-install-freshness.md
+++ b/.changeset/doctor-install-freshness.md
@@ -1,0 +1,24 @@
+---
+'nexus-agents': minor
+---
+
+teach `doctor` to notice when the global install has drifted from this build
+
+`.mcp.json` runs the MCP server off the **globally installed** package, not the
+working tree, and nothing keeps the two in step — the release workflow
+publishes to npm, the global install is a separate manual `npm install -g`.
+Drift accumulates silently, and every MCP call then executes code the operator
+is not looking at. Measured twice on 2026-08-25 alone: eleven minor versions
+behind in the morning, three more by the afternoon.
+
+`doctor` now compares the two and reports one of three states. `unknown` is its
+own state rather than folded into a pass: "no global install found" and "the
+versions match" are different facts, and the second is not something a check
+that could not measure is entitled to claim. `unknown` does not count as
+healthy, because not knowing is exactly the condition that let the drift
+accumulate.
+
+The remediation says to update **and restart any running MCP server**. An
+already-spawned server keeps the old code until it is restarted, so stopping at
+`npm install -g` reports the problem resolved while the session continues on
+the stale build.
--- a/packages/nexus-agents/src/cli/doctor-formatting.test.ts
+++ b/packages/nexus-agents/src/cli/doctor-formatting.test.ts
@@ -210,6 +210,7 @@ describe('doctor-formatting', () => {
         mismatch: false,
         dataDirInsideRepo: false,
       },
+      installFreshness: { state: 'aligned' as const, version: '1.0.0' },
       harnessAlignment: {
         agentsMdExists: true,
         files: [],
--- a/packages/nexus-agents/src/cli/doctor-install-freshness.test.ts
+++ b/packages/nexus-agents/src/cli/doctor-install-freshness.test.ts
@@ -1,0 +1,132 @@
+/**
+ * Tests for the doctor install-freshness sub-check (#4767).
+ *
+ * @module cli/doctor-install-freshness.test
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+  assessInstallFreshness,
+  describeInstallFreshness,
+  installFreshnessIsHealthy,
+  readGlobalVersion,
+} from './doctor-install-freshness.js';
+
+describe('assessInstallFreshness (#4767)', () => {
+  it('reports aligned when the versions match', () => {
+    expect(assessInstallFreshness('4.17.0', '4.17.0')).toEqual({
+      state: 'aligned',
+      version: '4.17.0',
+    });
+  });
+
+  it('reports behind with both versions named', () => {
+    // The operator needs both numbers to know how far they have drifted —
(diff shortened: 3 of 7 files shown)


williamzujkowski:
Producer/consumer ratchet flagged `checkInstallFreshness` — exported from `doctor.ts` and consumed only inside it. Unexported.

Nothing outside needs it. It is the `npm ls -g` call and nothing else; the verdict it wraps lives in `doctor-install-freshness.ts`, where it is pure, exported, and consumed cross-file by `doctor.ts` itself. Exporting the wrapper too would have created an export whose only importer is a test — which is exactly what the ratchet rejects, and it was right.

Fourth time today this check has caught a same-file-only export of mine, after `auditLineFor`, `partitionWarmStartSkips` and `visibleWidth`. The pattern in all four is identical: I extract a helper so it can be tested, export it for the test, and forget that a test importer is not a consumer. Worth noting that in three of the four the *right* answer was to move the logic to its own module, and here it is the opposite — keep the I/O private and test the pure part. The ratchet does not tell you which; it just refuses to let the question go unasked.

143 doctor tests pass.

williamzujkowski:
Third drift of the day, observed while this PR sat in CI:

```
global install: 4.15.5
npm / repo:     4.17.1
```

Two minors behind, roughly ninety minutes after I last synced it. The full sequence today:

| time | global | expected | gap |
| --- | --- | --- | --- |
| morning | 4.3.1 | 4.14.1 | 11 minors |
| midday | 4.14.1 | 4.15.2 | 1 minor |
| now | 4.15.5 | 4.17.1 | 2 minors |

Which is the argument for the check being *in `doctor`* rather than a one-off script or a note in the runbook. The gap re-opens every few releases and there is no moment at which anyone would think to look — I only caught this one because I happened to run `--version` while waiting on this PR's own CI.

It also makes the `unknown`-is-not-healthy decision concrete. On a machine where `npm ls -g` fails, the check would report `unknown`, and if that counted as a pass the operator would be in exactly the position I was in this morning: a green `doctor`, an MCP server eleven versions behind, and no signal anywhere.

Synced to 4.17.1.

Issue #4953, opened 2026-08-25
fix(governance+routing): two defects in today's own fixes — bandit vectors still differ, and a voided CLI vote records as rejected


williamzujkowski:
Two defects in code I merged today, found by an adversarial review of my own commits and verified against the tree. Both are the defect class the commit claimed to fix.

## 1. The bandit still trains on a different vector than it scores — on three columns, not one

#4911 fixed `budgetUtilization` and asserted the invariant in its own description: *"the feature vector handed to `update` must be the one `selectArm` scored."* That invariant is violated on **three of the six columns**, and was before my change.

The two paths use **different converters**:

| | select — `taskProfileToBanditContext` (`composite-router-stages.ts:666`) | update — `taskAnalysisResultToBanditContext` (`composite-router-outcome.ts:78`) |
| --- | --- | --- |
| `taskComplexity` | `profile.reasoningComplexity / 10`, and `reasoningComplexity = Math.round(score * 10)` → **quantized to 0.1** | `analysis.complexityScore` → **continuous** |
| `contextLengthNormalized` | `(estimatedTokens + 500) / 100000` — a documented legacy offset | `estimatedTokens / 100000` — **no offset** |
| `isReasoningTask` | `0` or `1` | `0`, `0.5`, or `1` |

`isReasoningTask` is the serious one: for a task with `complexityScore > 0.5` whose `reasoningType` is not `'reasoning'`, select supplies `0` or `1` while update supplies `0.5`. That is a half-unit error on a unit-scaled feature, an order of magnitude worse than the budget column I fixed.

The select path goes through `TaskProfile`, which exists to serve routing's 0-10 scale, and the update path skips it. Neither converter is wrong in isolation; using both is.

**Why my fix did not catch it:** the tests I added compare `budgetUtilization` only. Nothing asserts the vectors are equal, which is the invariant that actually matters — and an equality test over the whole context is the obvious thing to have written.

## 2. A policy-voided vote from the CLI records as `rejected`

#4925 gave the CLI an audit record. The MCP writer passes `errorVoided`:

```ts
errorVoided: result.policyReason !== undefined,   // consensus-vote.ts:856
```

My CLI writer passes four fields and not that one (`vote-command.ts:406-411`). `outcomeToDecision` (`audit/vote-record-store.ts:118-122`) then reads `errorVoided` as `false`:

```ts
if (errorVoided || (!result.quorumReached && allErrors)) return 'no_quorum';
return 'rejected';
```

[...]

PR #4989, opened 2026-08-26
fix(governance): record the resolved vote decision, not a second derivation

### LINKED ISSUE #4986: p1(governance): an absolute_quorum void records 'approved' in the audit chain while the tool reports no_quorum
## What

When `errorPolicy: 'absolute_quorum'` degrades a vote to `no_quorum`, the tool reports `no_quorum` and the **hash-chained audit record persists `approved`**. Both the CLI and MCP writers are affected.

A human spot-checking the chain sees an approval that did not happen. Per the mission section this is a p1 on the governor path, not a reporting nit: the record is the artifact the spot-check trusts.

## The trace

`outcomeToDecision` (`src/audit/vote-record-store.ts:113-123`) returns before it ever consults `errorVoided`:

```ts
if (result.outcome === 'approved') return 'approved';
const errorCount = votes.filter((v) => v.source === 'error').length;
const allErrors = errorCount === votes.length && errorCount > 0;
if (errorVoided || (!result.quorumReached && allErrors)) return 'no_quorum';
return 'rejected';
```

So `errorVoided` can only ever convert a `rejected` into a `no_quorum`. An `approved` outcome is unreachable by the void logic.

Meanwhile the response layer decides `no_quorum` *without* touching `result`:

1. `computeAbsoluteQuorumDecision` (`consensus-vote-types.ts:629`) returns `{decision:'no_quorum', degradeReason}` for two reachable cases — an errored voter (`:634`), and approved-by-responders below the absolute approval floor, the abstain-heavy case at `:653-657`.
2. `applyAbsoluteQuorumTelemetry` (`:694`) stamps that reason onto **`response`** only: `response.policyReason ??= degradeReason` (`:706`). `result.policyReason` stays `undefined`.
3. Both record writers derive the flag from the untouched field:
   - `consensus-vote.ts:854` — `errorVoided: result.policyReason !== undefined`
   - `cli/vote-command.ts:420` — same expression

Result: `errorVoided === false`, `result.outcome === 'approved'`, record `decision: 'approved'`, response `decision: 'no_quorum'`.

## Reachability

Both entry points, verified against main today:

- **MCP** — `errorPolicy` is in the tool schema.
- **CLI** — `--error-policy absolute_quorum` is forwarded: `cli-commands-handlers.ts:326-332` validates it into the options, `vote-command.ts:312` puts it on the `ConsensusVoteInput`. (An earlier read of mine said the CLI could not reach this; that was true before #4964 landed this morning and is no longer.)

## Failing assertion

```
recordAuthenticVote({
  result: { outcome: 'approved', quorumReached: false, ... },
  votes: <no errored voters>,
  errorVoided: false,
})
```
→ persisted `.decision === 'approved'`, while `resolveVoteDecision` on the same result returns `no_quorum`.

## Why the existing guard does not catch it

#4953 fixed this exact class in the other direction — a voided vote recorded as `rejected` — and placed the `errorVoided` check *after* the `approved` short-circuit. The guard therefore rescues only the rejection half of the problem it was filed for. The comment above it claims the record and the response "never disagree"; they do, in one direction, and nothing tests that direction.

## Suggested fix

Stop re-deriving the decision in the audit layer. `resolveVoteDecision` already computes the authoritative three-valued answer once, and `ExtendedVotingResult.decision` carries it (#4135) — pass it into `buildVoteRecord` and record it, keeping `outcomeToDecision` only as the fallback for callers that have no resolved decision. That removes the second derivation rather than patching it, which is what let the two disagree.

Add the missing-direction test: an `approved` outcome with a `no_quorum` resolution must not persist as `approved`.

## Governance

`src/audit/` is a governance path — owner ratification required, never auto-merged. Filing rather than self-merging.

## Provenance

Surfaced by an adversarial review agent over eight of today's self-merged commits; the trace above was re-verified by hand against main, and one of the agent's scope claims was wrong and is corrected inline.

williamzujkowski:
Closes #4986.

**Owner ratification required — touches `src/audit/`. I will not self-merge this.**

## The misreport

`outcomeToDecision` returned before it ever consulted `errorVoided`:

```ts
if (result.outcome === 'approved') return 'approved';   // <- short-circuit
...
if (errorVoided || (!result.quorumReached && allErrors)) return 'no_quorum';
```

So `errorVoided` could only ever convert a `rejected` into a `no_quorum`. Meanwhile `computeAbsoluteQuorumDecision` resolves `no_quorum` and `applyAbsoluteQuorumTelemetry` stamps the reason on **`response`** only (`consensus-vote-types.ts:706`) — `result.policyReason` stays `undefined`, which is exactly the field both writers use for `errorVoided`.

Response: `no_quorum`. Hash-chained record: `approved`. Reachable from MCP (`errorPolicy` is in the schema) and from the CLI (`--error-policy absolute_quorum`, forwarded since #4964).

#4953 fixed this same class in the other direction and placed its guard below the short-circuit, which is how the second half survived. The comment above it says the record and response "never disagree."

## The fix is to delete the second derivation, not patch it

`resolveVoteDecision` already computes the authoritative three-valued answer once, and `executeVoting` stamps it on the result (`consensus-vote.ts:621`). The record now takes that. `outcomeToDecision` remains only as the fallback for callers that genuinely have no resolved decision.

`resolvedDecision` is **required**, including its `undefined` case. Optional would have let a future call site inherit the fallback in silence — the exact failure mode being fixed. Making it required turned this into 13 compile errors naming every call site, which is the enumeration a test would have had to guess at.

## Two adjacent misreports found while fixing it

**A voided approval, in the fallback too.** The `errorVoided` check moved above the `approved` short-circuit, so the fallback is no longer asymmetric.

**A timed-out panel recorded as a rejection.** `ConsensusResult.outcome` is `ProposalStatus` — six values, including `timeout`, `pending`, `voting`, `closed`. The old `return 'rejected'` default recorded every one of them as the panel voting the proposal down, attributing a verdict to voters who never gave one. Now only `rejected` is a rejection; the rest are `no_quorum`.

The response vocabulary is wider than the record's, so the narrowing is stated explicitly in `toRecordDecision` rather than cast: `timeout` → `no_quorum`, `pending` → `undefined` (fall back; a vote in flight has no decision to record).

## Mutations

| mutation | result |
| --- | --- |
| ignore `resolvedDecision`, always derive | 1 failed |
| restore the original check ordering | 1 failed |
| `timeout` back to `rejected` | 1 failed |

A paired test keeps the third change from swallowing genuine rejections. 7,183 tests pass across `src/audit`, `src/cli`, and the consensus-vote suites. `api-surface.txt` is unchanged — none of these are public exports.

## Provenance

Surfaced by an adversarial review agent over eight of today's self-merged commits, then re-traced by hand. One of the agent's supporting claims — that the CLI could not reach `absolute_quorum` — was true before #4964 landed this morning and is corrected in the issue.

### DIFF
--- a/.changeset/audit-record-quorum-void.md
+++ b/.changeset/audit-record-quorum-void.md
@@ -1,0 +1,24 @@
+---
+'nexus-agents': patch
+---
+
+fix(governance): stop the audit chain recording a voided vote as approved
+
+`outcomeToDecision` returned on `result.outcome === 'approved'` before it
+consulted `errorVoided`, so the void logic could only ever convert a `rejected`
+into a `no_quorum`. An `absolute_quorum` degradation stamps its reason on the
+response and never on the result, so `errorVoided` was false, and the
+hash-chained record persisted `approved` for a vote the tool reported as
+`no_quorum`.
+
+The record now takes the decision `resolveVoteDecision` already produced, rather
+than deriving a second one from the outcome — the second derivation is what let
+the two disagree. `resolvedDecision` is a required field, including its
+`undefined` case, so the compiler names every call site instead of letting a new
+one inherit the fallback silently.
+
+Two adjacent misreports went with it. The fallback's `errorVoided` check moved
+above the `approved` short-circuit, and a non-terminal outcome is no longer
+recorded as a rejection: `ProposalStatus` also carries `timeout`, `pending`,
+`voting` and `closed`, and the old everything-else-is-rejected default
+attributed a verdict to voters who never gave one.
--- a/packages/nexus-agents/src/audit/vote-record-store.test.ts
+++ b/packages/nexus-agents/src/audit/vote-record-store.test.ts
@@ -78,6 +78,8 @@ const votes: readonly AgentVoteResult[] = [
 describe('buildVoteRecord', () => {
   it('carries the proposal hash, decision, counts, per-voter summary, and a sequence', () => {
     const record = buildVoteRecord({
+      // #4986: these fixtures exercise the fallback derivation.
+      resolvedDecision: undefined,
       id: 'vote-1',
       proposal: 'Promote loop X to enforce',
       strategy: 'higher_order',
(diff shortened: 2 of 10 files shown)


williamzujkowski:

PR #5015, opened 2026-08-26
fix(observability): stop a non-auditable fitness result emitting a below-floor signal

### LINKED ISSUE #5014: improvement_review emits a below-floor fitness signal for an audit that could not run
Found auditing verdict fidelity in the review/gate tools. Fix in flight.

## The asymmetry

`emitFitnessDeclinedSignal` (`mcp/tools/improvement-review-signals.ts:41`) gated only on the score:

```ts
if (audit.score >= fitnessFloor) return;
```

Its sibling `detectFitnessSignals` (`mcp/tools/improvement-review.ts:667`) has refused the non-auditable case since #3621:

```ts
// #3621: a non-auditable result (e.g. run from the global npm install) has a
// meaningless score of 0 — it is "could not audit", not "fitness is low".
if (audit.auditable === false) return signals;
```

`improvement-review.ts:1027` passes **the same audit object** to both.

## What it produced

`improvement_review` run through the global npm install — which is the configured `.mcp.json` path — hits `calculateFitnessScore`'s not-source-repo sentinel (`{ score: 0, auditable: false }`, `governance/fitness-score.ts:337`) because the bundled `src/` lacks `cli-adapters/`. The emitter then published `signal.fitness_declined { score: 0, floor: 90 }` onto the pipeline bus, and `tune-stage` turned that into a `flag_tech_debt` reading "fitness 0 below floor 90" — for a repo nobody audited.

The response payload carries no fitness field at all, so the fabricated signal was invisible at the tool boundary. The `auditable` doc comment says consumers "MUST NOT treat a `false` audit's score as a signal"; one consumer did.

## Fix

The rule is now one exported predicate, `isAuditableScore`, called by both. A second copy of the guard would have fixed today's case and left the next consumer free to miss it the same way.

`improvement-review-signals.test.ts` had no `auditable` case; it has two now — the sentinel emits nothing, and a genuine below-floor audit still emits.

## A correction worth recording

Mid-investigation I grepped for `auditable` across the tree, found no consumer outside the type definition, and concluded that **nothing** guarded this — a strictly worse finding than the real one. That was wrong: I had already replaced `audit.auditable === false` with `!isAuditableScore(audit)` in the sibling, so my own edit had deleted the string I was grepping for. The compiler caught it (the import was missing) before it reached an issue. Worth noting because "grep finds nothing" is only evidence when the tree is clean.

williamzujkowski:
Closes #5014.

## One audit object, two consumers, one guard

`emitFitnessDeclinedSignal` (`improvement-review-signals.ts:41`) gated only on `audit.score >= fitnessFloor`. Its sibling `detectFitnessSignals` (`improvement-review.ts:667`) has refused `auditable === false` since #3621, with the reason written down: the not-source-repo sentinel carries a meaningless score of 0, and that is *"could not audit"*, not *"fitness is low"*. `improvement-review.ts:1027` hands both functions the same object.

So `improvement_review` run through the global npm install — the configured `.mcp.json` path, where the bundled `src/` lacks `cli-adapters/` — emitted `signal.fitness_declined { score: 0, floor: 90 }`, and `tune-stage` turned it into a `flag_tech_debt` reading "fitness 0 below floor 90" for a repo nobody audited. The response payload has no fitness field, so nothing at the tool boundary disclosed it.

## Hoisted, not copied

A second copy of the guard fixes today's case and leaves the next consumer free to miss it identically. The rule is now one exported predicate, `isAuditableScore`, called by both — which is also where the `auditable` field's contract ("consumers MUST NOT treat a `false` audit's score as a signal") can actually live.

| mutation | result |
| --- | --- |
| drop the emitter guard | 1 failed |
| emitter never emits | 6 failed |

The second is the pair: refusing the sentinel must not silence genuine low fitness. `improvement-review-signals.test.ts` had no `auditable` case at all; it has both directions now.

52 tests pass across the improvement-review suites. `api-surface.txt` unchanged.

## A wrong turn worth recording

Mid-investigation I grepped for `auditable` across the tree, found no consumer outside the type definition, and started writing up a strictly worse finding: that **nothing** guarded this. It was wrong — I had already rewritten the sibling's `audit.auditable === false` into `!isAuditableScore(audit)`, so my own edit had removed the string I was searching for. The compiler caught it (missing import) before it reached the issue. "Grep finds nothing" is only evidence when the tree is clean.

### DIFF
--- a/.changeset/fitness-signal-unauditable.md
+++ b/.changeset/fitness-signal-unauditable.md
@@ -1,0 +1,20 @@
+---
+'nexus-agents': patch
+---
+
+fix(observability): stop a non-auditable fitness result emitting a below-floor signal
+
+`emitFitnessDeclinedSignal` gated only on `audit.score >= fitnessFloor`. Its
+sibling `detectFitnessSignals` has refused `auditable === false` since #3621 —
+the not-source-repo sentinel carries a meaningless score of 0, which is "could
+not audit", not "fitness is low" — and both are handed the same audit object.
+
+So `improvement_review` run from the global npm install, where the bundled
+`src/` lacks `cli-adapters/`, emitted `signal.fitness_declined { score: 0,
+floor: 90 }` onto the pipeline bus, and `tune-stage` turned that into a
+"fitness 0 below floor 90" tech-debt flag for a repo that was never audited. The
+response payload carries no fitness field, so the fabricated signal was
+invisible at the tool boundary.
+
+The rule now lives in one exported predicate, `isAuditableScore`, that both
+consumers call, so they cannot drift apart again.
--- a/packages/nexus-agents/src/governance/fitness-score.ts
+++ b/packages/nexus-agents/src/governance/fitness-score.ts
@@ -172,6 +172,21 @@ export interface FitnessAudit {
 }
 
 /**
+ * Whether an audit's score is a real measurement rather than the
+ * "could not audit from here" sentinel (#3621).
+ *
+ * Exported as one predicate because the guard was applied by
+ * `detectFitnessSignals` and missed by `emitFitnessDeclinedSignal`, which was
+ * handed the same audit object and emitted `signal.fitness_declined` with the
+ * sentinel's score of 0 — a tech-debt signal claiming catastrophic fitness for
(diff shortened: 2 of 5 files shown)


williamzujkowski:

Issue #5028, opened 2026-08-26
CI gates that cannot fail: a live [skip-docops] bypass, a ratification gate with no producer, and three more


williamzujkowski:
Found auditing the CI gate scripts for the "check that cannot fail" class, scoped to gates #4586 does not cover. Fix in flight for finding 1; the rest are filed here.

## 1. The DocOps gate can be bypassed today by branching from an old commit

`scripts/check-docops-skill.ts:163` reads:

```ts
execFileSync('git', ['log', `origin/${baseRef}...HEAD`, '--pretty=%B'], …)
```

Three dots is the **symmetric** difference, so it also returns commits reachable from the base branch but not from HEAD — including every `[skip-docops]` ever merged to `main`. Six are there today, the most recent `96f63f7e74` (2026-07-06).

Any PR branch whose merge-base predates one of them has it in the range, so `runCheck` takes the `escapeHatchUsed = true; success = true` early return at `:258` and exits 0 with the whole skill-sync gate skipped — regardless of what the PR changed.

**Demonstrated on this repo**, using a commit that predates the markers as a stand-in HEAD:

```
three-dot (current code): 7 occurrences of [skip-docops]
two-dot  (correct):       0
```

The commit that introduced this range — `919de90479`, *"fix(ci): [skip-docops] honored on PR runs"* — is itself one of the six.

## 2. The authority-tier ratification gate has no producer

`scripts/check-authority-tier-drift.ts:482` — `ratificationGateFindings()` reads `governance/authority-tier-transitions.jsonl`, which is **0 bytes and has no writer in the tree**. `logTierTransition` (`audit-logger.ts:535`) is called only from tests and no-op stubs. Empty file → no transitions → `analyzeTierTransitionEvents([])` → zero findings, always.

So the #3842/#3894/#3927 path that resolves a `ratificationVoteRef` against the tamper-evident `vote-records.jsonl` cannot emit a finding. What actually guards a promotion is `:463`, `if (e.ratificationVote.trim() === '')` — the cosmetic non-empty check #3842 says it replaced.

**Scenario:** flip a manifest's `authorityTier` to `enforce` and add an evidence entry with `ratificationVote: "#9999"` naming a vote that does not exist. The gate exits 0. This is the governor path.

## 3. Every knip failure reports zero orphans

[...]

Issue #80, opened 2026-01-04
feat(mesh): implement mode auto-detection


williamzujkowski:
## Overview
Implement automatic detection of nexus-agents invocation mode based on environment and stdin.

**Phase:** 3 (Hybrid Mesh)
**Target Version:** v3.0.0
**Epic:** #70
**Depends on:** #73

### Mode Detection Logic
```typescript
function detectMode(): 'server' | 'orchestrator' | 'mesh' {
  // Check for explicit --mode flag first
  const explicit = args.mode;
  if (explicit) return explicit;

  // Check if stdin is a TTY
  const isInteractive = process.stdin.isTTY;
  
  // Check if we're being invoked by an MCP client
  const mcpClient = process.env.MCP_CLIENT_NAME;
  
  // Check for CI/non-interactive environment
  const isCI = process.env.CI === 'true';

  if (mcpClient || !isInteractive) {
    return 'server'; // We're being called as MCP server
  }
  
  if (isCI) {
    return 'orchestrator'; // CI runs, we orchestrate CLIs
  }

  return 'mesh'; // Interactive, full capabilities
}
```

### Mode Behaviors

| Mode | nexus-agents As | Can Call |
|------|-----------------|----------|
| server | MCP server only | N/A (responds to calls) |
| orchestrator | CLI orchestrator | Claude, Gemini, Codex CLIs |
| mesh | Both | All + can be called by Claude |

## Acceptance Criteria

- [ ] TTY detection for interactive mode
- [ ] MCP client detection from environment
- [ ] CI environment detection
- [ ] Explicit --mode flag overrides auto-detection
- [ ] Mode logged at startup
- [ ] <100ms detection time
- [ ] Unit tests for all detection scenarios
- [ ] Integration tests for mode switching

## Files to Modify

- `src/cli.ts` - Add detection logic
- `src/mcp/server.ts` - Mode-aware initialization

---
*Part of CLI Integration Phase 3 per cli-project_plan.md v2.0.0*

Issue #5022, opened 2026-08-26
p1(security): ClawGuard is behaviorally identical to 'off' in every mode — the policy is never in scope at dispatch, and allowedTools is always empty


williamzujkowski:
Traced end to end: enforcement point → middleware mount → ALS producer → production callers. **In every mode, ClawGuard is behaviorally identical to `off`.**

Unlike the `PolicyFirewall` case (#4888), this is not a barrel-export-and-tests artifact. The middleware is genuinely mounted, derivation genuinely runs, a durable audit sink genuinely exists. The two halves just never overlap in async scope.

## 1. The guard is mounted on every tool, but its policy is never in scope at dispatch

**Enforcement half — wired.** `buildMiddlewareStack` unconditionally pushes `createAccessPolicyChainMiddleware(config.toolName)` (`mcp/middleware/middleware-chain.ts:355-358`), reached by real tools via `wrapToolWithTimeout` → `withMiddleware` (`tool-wrapper.ts:203`). ~28 registered tools go through it.

**Policy half — wired, but elsewhere.** `withAccessPolicy` has exactly two non-deriver callers, and neither wraps MCP tool dispatch:

- `mcp/tools/orchestrate.ts:743` wraps `() => orchestrator.execute(definition, {})`
- `mcp/tools/execute-expert.ts:616` wraps `() => expert.execute(task)`

**The gap.** `AsyncLocalStorage` propagates only into async work spawned *inside* `run()`. A registered tool handler is invoked by an inbound MCP request from the SDK transport — a sibling async context, never a descendant of an in-flight `orchestrate` call. So `getActivePolicy()` (`chain-adapter.ts:36`) returns `undefined` and line 38 returns `next(args, ctx)` unconditionally.

The guard could only fire if orchestrator/expert execution dispatched a *wrapped* handler in-process. It does not: the in-process call sites import the **inner** business functions and bypass the chain — `pipeline/agent-executor.ts:24` (`executeDiscovery`), `:25` (`analyzeGaps`), `:453` (`generateWeatherReport`), `:629` and `pipeline/iterative-consensus.ts:269` (`executeVoting`).

Note `orchestrate` is itself a wrapped tool (`:1478`), so the ClawGuard middleware runs on the `orchestrate` call *before* its handler establishes the policy — with `undefined` in ALS.

[...]

PR #5033, opened 2026-08-26
fix(governance): make the authority-tier ratification gate able to fire

williamzujkowski:
Closes finding 2 of #5028. Remedy chosen by a 7-voter panel: **Option C**, 5 of 5 approvers, 83.3%, audit record #80.

**Owner ratification required — touches `governance/`. I will not self-merge.**

## The gate had no input

`ratificationGateFindings()` read `governance/authority-tier-transitions.jsonl`: **0 bytes**, created 2026-06-16, never written. Its only writer, `IAuditLogger.logTierTransition`, is called from nothing but tests and a no-op stub. Empty file → no transitions → `analyzeTierTransitionEvents([])` → zero findings, unconditionally.

So the whole #3842/#3894/#3927 path that resolves a `ratificationVoteRef` against the tamper-evident `vote-records.jsonl` could not emit a finding. What actually guarded a promotion was `if (e.ratificationVote.trim() === '')` — a non-empty-string test, which is the cosmetic check #3842 states it replaced.

## Why no producer was ever written

**Authority tiers are not changed at runtime.** They live in `governance/loop-tiers.yaml`, a checked-in file, and the only commit that has ever changed one is `8ca02e8879`. A tier transition is a *commit*, not a runtime event — so `logTierTransition` was waiting for something that does not happen. That fact is what made the panel's three options genuinely different, and it is why the answer is not "go write the missing producer."

The evidence ledger, by contrast, exists and already carries `ratificationVote`. The resolution machinery is now pointed there.

## Verified end to end, not just in unit tests

With a promotion entry citing `ratificationVote: '#9999'`:

```
exit=1
[promotion-ratification-unresolved] tier-transition promotion of 'dev-pipeline'
(advisory → enforce) carries ratificationVoteRef='#9999' that does NOT resolve
to any authentic record in governance/vote-records.jsonl.
```

With the ledger restored: `exit=0`. That is the pair, and it is the scenario that previously exited 0.

| mutation | result |
| --- | --- |
| gate returns nothing | 2 failed |
| drop the collected findings | 1 failed |

51 tests pass.

## Two things I deliberately did not do

**`logTierTransition` stays.** It is in `api-surface.txt`, so removing it is a breaking API change needing its own unanimous vote — and #5003 showed that bar is real (a removal at 83.3% was rejected). It is now dead weight with a documented reason; removing it belongs in a separate, correctly-gated change.

**The positive unit test asserts fail-closed, not pass.** My first attempt hand-wrote a vote record to prove a valid ratification clears the gate; it failed, because the resolver runs tamper-evidence verification and a forged hash correctly fails the whole ledger closed. Rather than fabricate a valid hash chain to make a test go green, the unit test now pins that behaviour — an unverifiable ledger rejects every promotion — and the passing direction is covered by the end-to-end run above.

### DIFF
--- a/.changeset/ratification-resolves-vote-record.md
+++ b/.changeset/ratification-resolves-vote-record.md
@@ -1,0 +1,22 @@
+---
+'nexus-agents': patch
+---
+
+fix(governance): make the authority-tier ratification gate able to fire
+
+The gate read `governance/authority-tier-transitions.jsonl` — a 0-byte file
+whose only writer, `IAuditLogger.logTierTransition`, is called from nothing but
+tests and a no-op stub. Empty file → no transitions → zero findings, always. So
+the #3842/#3894/#3927 path that resolves a `ratificationVoteRef` against the
+tamper-evident `vote-records.jsonl` could not emit a finding, and the only live
+guard on a promotion was a non-empty-string test — the cosmetic check #3842
+states it replaced. A manifest could be promoted to `enforce` citing a vote that
+does not exist.
+
+Authority tiers are not changed at runtime: they live in
+`governance/loop-tiers.yaml` and change by commit, so the runtime event the
+design waited for never occurs. The resolution machinery is now pointed at the
+promotion-evidence ledger, which does exist and does carry `ratificationVote`.
+The dead transitions file is removed.
+
+Remedy chosen by a 7-voter panel: Option C, 5 of 5 approvers, audit record #80.
--- a/scripts/check-authority-tier-drift.test.ts
+++ b/scripts/check-authority-tier-drift.test.ts
@@ -13,6 +13,7 @@
 import { describe, it, expect } from 'vitest';
 import { stringify as toYaml } from 'yaml';
 import {
+  ratificationGateFindings,
   analyzeTierDeclarations,
   analyzeLoopTierDeclarations,
   analyzeTierTransitionEvents,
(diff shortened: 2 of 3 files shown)


williamzujkowski:
## Before ratifying: the ledger this repoints to is also empty

Found while auditing producer/consumer seams. **This PR is a real improvement and I am not arguing against merging it** — it removes the cosmetic non-empty-string check and adds genuine resolution. But it will not make the gate able to fire in production, and the PR body reads as though it will.

### Measured on `main` today

```
governance/vote-records.jsonl              0 bytes   (unchanged since 2026-06-17)
governance/pr-review-records.jsonl         0 bytes   (unchanged since 2026-08-15)
governance/authority-tier-transitions.jsonl 0 bytes  (unchanged since 2026-08-26)

.nexus-agents/governance/vote-records.jsonl  144,103 bytes, 91 records
                                             last written 2026-08-27 15:10
```

Those 91 records include today's `higher_order` panels — the #5022 ClawGuard decision among them. **Vote records are being written. They are being written somewhere the gate does not read.**

### Why

`vote-record-store.ts:59-65` states it plainly:

> As of #3991 the runtime store no longer auto-writes here — this committed artifact is reached **only via the `NEXUS_VOTE_RECORDS_PATH` override** (the separate caller-commits/governor-gate path, **deferred**).

`resolveVoteRecordsPath()` (`:351`) takes the env override if set, otherwise `nexusDataPath('governance', 'vote-records.jsonl')` — the gitignored runtime dir.

**Nothing sets that override.** `grep -rn NEXUS_VOTE_RECORDS_PATH .github/ scripts/ package.json` returns nothing.

So the "caller-commits" bridge was deliberately deferred in #3991 and never filed. The deferral is documented in a JSDoc; there is no issue tracking it.

### What that means for this PR

The finding this closes is *"the gate read a 0-byte file, so it aggregated over nothing."* After this change it reads a **different** 0-byte file. The gate becomes able to fire — genuinely, and the mutation matrix proves the code path — but its input still never arrives.

The PR's own end-to-end verification says so without quite saying so:

> With the ledger restored: `exit=0`.

"Restored" is the tell. That is the ClawGuard shape exactly: the test supplies the one input production never supplies. `access-constraint-deriver` had 124 passing tests over inert code for the same reason.

### Recommendation

Ratify this — it is strictly better than what it replaces, and the resolution machinery is correct and tested. But do not treat it as closing the promotion-gate hole, and do not let the changeset say the gate now fires. The remaining work is the deferred bridge: something must publish authentic vote records to the committed ledger, or the gates must read the runtime one.

Filing that as its own issue now, since a JSDoc deferral is not tracking. It also affects `check-governor-review.ts`, which reads the equally-empty `governance/pr-review-records.jsonl` and therefore takes its warn-first "no diff-bound record" branch, exit 0, on every governor PR.

### Not in doubt

`check-governor-ratification.ts` — the gate currently red on all six open governor PRs — reads **no** ledger; it resolves via CODEOWNERS approval or the label. That one works. I measured it separately today (6 of 6, only check failing). The ledger problem is confined to the gates that consume the record, not the one that gates the merge.

williamzujkowski:

Issue #5099, opened 2026-08-27
test-hygiene: ast-rule-runner.test.ts is cwd-dependent — red from the repo root, green from the package dir


williamzujkowski:
## What

`ast-rule-runner.test.ts` fails or passes depending on the **working directory** the runner is invoked from, because one assertion builds a fixture path relative to `process.cwd()`.

```
FAIL src/security/ast-rule-runner.test.ts > unreadable / non-directory scan root fails LOUD — #4277 gap 2
     > throws when targetDir is a FILE, not a directory (no empty "clean" result)

AssertionError: expected [Function] to throw error matching /must be a directory/
  Received: "ast-rule-runner: cannot read targetDir
             <repo>/src/security/ast-rules/fixtures/sample.py:
             ENOENT: no such file or directory"
```

## Reproduce

```bash
# from the repo root — FAILS (1 failed | 16 passed)
npx vitest run --root packages/nexus-agents src/security/ast-rule-runner.test.ts

# from the package dir — PASSES (17 passed)
cd packages/nexus-agents && npx vitest run src/security/ast-rule-runner.test.ts
```

## Cause

`ast-rule-runner.test.ts:299` builds the path with `join('src', 'security', 'ast-rules', 'fixtures', 'sample.py')` — no root anchor. Vitest's `--root` sets the project root for *resolution*, not `process.cwd()`, so from the repo root the relative path resolves to `<repo>/src/...` instead of `<repo>/packages/nexus-agents/src/...`. The fixture is absent there, so `runAstQaRules` throws ENOENT before reaching the is-a-directory check the assertion targets.

## Severity: low, and the security property still holds

The production path is fine — `runAstQaRules` is called with an absolute or caller-supplied `targetDir`, and this is a **test-only** cwd dependency. The #4277 gap-2 property being asserted (fail LOUD, never return an empty `{findings: []}` that reads as clean) is **still satisfied** in the failing case: it throws, just with a different message. So this is test hygiene, not a regression in the guard.

It is worth fixing anyway for two reasons:

1. It makes `--root`-style invocations red for an unrelated reason, which is exactly the noise that trains people to ignore a red suite.
2. The assertion currently passes for a *narrower* reason than it claims. It asserts `/must be a directory/`, but from the package dir it happens to reach that branch; nothing pins that the ENOENT branch and the not-a-directory branch are distinct. A fix should anchor the path **and** add the missing-path case as its own assertion, so the two failure modes cannot collapse into each other again.

[...]

Issue #4888, opened 2026-08-25
security(mcp): PolicyFirewall is constructed and never reaches any tool's middleware chain


williamzujkowski:
Split from a security-ledger finding. The **false claim** is fixed separately (the startup log no longer reports enforcement); this issue is the actual wiring.

## Verified against main

`createConfiguredPolicyFirewall` builds a firewall (`cli-server-audit.ts:124`), `cli-server.ts:517` receives it, and `registerMcpTools` destructures it (`cli-server-tools.ts:827`).

**It is then used for exactly one thing: a log line.** `buildStandardDeps` (`:600-618`) returns `logger`, `rateLimiter`, `security` and `auditLogger` — never `policyFirewall`. `grep -n policyFirewall src/mcp/tools/*.ts` returns zero non-test hits.

The support exists and is unreached: `middleware-chain.ts:324` adds the policy middleware `if (skip.policy !== true && config.policyFirewall !== undefined)`, and `createPolicyMiddleware` (`:151`) has exactly one caller — that line. With nothing ever supplying the config field, **no policy middleware runs for any registered tool.**

## Observable failure

An operator sets a policy mode, the rules load (`policyRuleCount` is logged from `getRules().length`, so they are real), and every tool call proceeds without consulting them. Nothing denies, and until the companion fix nothing said so.

## Why this is filed rather than fixed

The wiring is not a one-liner and it carries a decision:

1. **Scope.** Forwarding through `buildStandardDeps` reaches the tools that use it, but not every registration goes through that path — the ~47 tools register via several shapes, and some pass bespoke deps.
2. **Enforcement mode is a live behaviour change.** Turning on a firewall that has never run means rules never exercised against real traffic start denying. That wants the same staged treatment other gates in this repo got: audit → soak → enforce, not a flip.
3. **`skip.policy`** already exists in the chain config, so per-tool exemptions are expressible — but nothing decides which tools warrant one.

## Suggested shape

Wire it in audit mode first: policy middleware attached everywhere, decisions recorded, nothing denied. That produces the evidence for an enforce decision instead of asking for it up front — and it matches how `NEXUS_ACCESS_POLICY_MODE` and `NEXUS_AUTO_REMEDIATE` were staged.

## Trigger

Pick up when someone wants policy enforcement to actually apply, or alongside the next access-policy work. Until then the honest state is a warning at startup saying the firewall is inert.

Related: the companion fix removing the false `policyFirewallEnabled: true` / `policyMode` log fields.

Issue #2077, opened 2026-04-20
feat(security): consider flipping ClawGuard default from audit to enforce after bake period


williamzujkowski:
Follow-up to #2064 which flipped \`NEXUS_ACCESS_POLICY_MODE\` default from \`off\` to \`audit\` in v2.50.

## Current state

- **Default:** \`audit\` — every orchestrate/execute_expert call derives a policy and logs \`access-policy: audit violation\` when tool calls fall outside the derived allowlist
- **Block mode** (\`NEXUS_ACCESS_POLICY_MODE=enforce\`) requires opt-in

## Proposed next step (not for v2.51; after a bake period)

Review \`access-policy: audit violation\` log frequency over 1-2 weeks of v2.50.x usage. If:
- Violation rate is acceptably low (< ~5% of tool calls)
- False-positive rate is near-zero (violations represent real risk)
- No critical-path tool categories are consistently flagged

…then flip the default to \`enforce\`.

## Operational prerequisites

- Structured telemetry collection (partially there via \`access-policy\` log entries)
- An opt-out path documented for users who need \`audit\` or \`off\` (already in CLAUDE.md env-var table)
- Regression test that verifies a user can always downgrade via env var

## Decision gates

This is a **user-visible behavior change** — orchestrations that previously succeeded would now block on policy violations. It should wait for:
1. Explicit product decision (owner: @williamzujkowski)
2. Evidence that audit-mode telemetry shows manageable violation rate
3. A minor-version bump (v2.x → v2.y, not a patch)

## Why not flip now

Zero production telemetry data exists yet. Flipping prematurely risks breaking real workflows. \`audit\` mode specifically exists to gather that data first.

Related: #1977, #2024, #2064.

williamzujkowski:
Re-verified 2026-06-08: ClawGuard default is still `audit` (`security/access-constraint-deriver/config.ts:20`: `DEFAULT_ACCESS_POLICY_MODE = 'audit'`). (Note: `SECURITY_DEFAULTS.policyMode='enforce'` in `config/defaults.ts:310` is the unrelated pipeline policy-execution mode, not ClawGuard.) Decision gates unmet: no product decision logged, and the structured audit-violation telemetry needed to judge the <5% violation / near-zero false-positive bar isn't yet aggregated. Trigger has not fired — keeping open as the tracked deferred decision. Still p2.

williamzujkowski:
Closing as **superseded**, not done.

[...]

PR #5100, opened 2026-08-27
fix(security): report an empty ClawGuard allowlist as unmeasured, not as a deny

### LINKED ISSUE #5022: p1(security): ClawGuard is behaviorally identical to 'off' in every mode — the policy is never in scope at dispatch, and allowedTools is always empty

williamzujkowski:
Partially addresses #5022. **Does not answer the boundary question** — see below.

## The vote did not produce a mandate

I put #5022's boundary question to a `higher_order` panel at the supermajority bar with the issue's four options. Result:

| | |
|---|---|
| Approve / reject | **6 / 1** (85.7%) — clears the bar |
| Option split | **B:2, C:2, A:1, D:1** |
| Leading share | **33.3%** — does not clear the bar |
| Verdict | `rejected`, `thresholdMet: false` |

So the panel agrees the defect is real and disagrees on where the guard belongs. I am not breaking a 2-2-1-1 tie on a security boundary. **This PR lands only what all four options require**, and leaves the boundary open.

Worth recording: 4 of the 7 voters — architect, ai_ml, catfish, scope_steward — independently raised the same objection, that inbound MCP dispatch is already owned by `PolicyFirewall` (#4888/#4987) and a second guard there would be a parallel implementation. That coalition is larger than any single option won, and it argues against A and C. It is the strongest signal in the record and belongs in the re-vote framing, not in a unilateral decision here.

## What all four options required

**1. An empty allowlist means `unmeasured`, not deny.**

`checkAccess` decides on one field. No production producer writes a tool name into it — the LLM deriver is asked for `tool_categories` / `file_scope` / `network_scope` and pins `[]`; the keyword fallback hardcodes `[]`; both derivation-failure paths pick between `[]` and `'*'`. So `[].includes(name)` was false for every call and the verdict was a constant function of `(mode, isRiskyTool(name))` — independent of the objective, the LLM output, and the trust tier.

A new `unmeasured` variant on `AccessDecision` says the allowlist arm did not run. That is not "examined and refused", and it is not "examined and passed". The adapter allows the call, logs it, and **deliberately does not record it as a violation** — recording a check that never ran would give the #2077 enforce-flip denominator a definitionally 100% violation rate carrying no information about precision. The unbypassable denylist still runs first and still denies.

This is also the change that makes the boundary question *safe to answer later*. Establishing a policy at dispatch without it would have denied every guarded call under `enforce`, and every non-read-only tool under `confirm_risky`.

**2. Delete the dead parallel adapter.** `guardMcpToolCall` and `createAccessPolicyMiddleware` had no production caller and no entry in `api-surface.txt`, so this is not a public API change. Their unique coverage — the #4097 audit-always-allows regression, including the throwing-sink case — is ported to the adapter that actually runs, not dropped.

**3. Correct the false documentation.** Three module headers described behaviour the code does not have: that `off` is the default (it has been `audit` since v2.50), and that the denylist protects `~/.ssh/**`, `~/.aws/**` and `/etc/shadow` "even in `off` mode". It is reachable only from `checkAccess`, which an absent policy short-circuits before.

**4. A reachability gate.** `access-policy-reachability.test.ts` runs a handler through the real middleware stack instead of establishing the policy itself, and records that no policy is in scope at inbound dispatch.

Also fixed, uncontested: a denial logged at `info` while an audit-mode observation logged at `warn` — the blocking mode logged *below* the observing one.

## Why the existing tests passed

**124 of them, across 9 files, over a subsystem that was a pass-through for every real dispatch.** Every one calls `withAccessPolicy` itself first. That is the one input production never supplies, so the suite could not fail for the reason production was broken. The new test is the inverse: it asserts reachability, not verdicts.

It pins the current *broken* behaviour on purpose, which I would normally not do. The rationale is in the file header: with the boundary question open, the failure mode to prevent is a **silent** change. If these go red, that is the signal to resolve #5022 — not to relax the assertion.

## Deliberately NOT in this PR

**The durability inversion.** `recordAuditModeViolation` is called only on the log-and-allow branch; the deny branch writes an ephemeral `logger.info`. The mode that only observes writes to the hash chain and the modes that block do not. Three of the four options want this fixed, but doing it properly means renaming the sink and extending the `clawguard_violation` event shape — which changes hash-chain *content* and deserves its own review. Filed separately rather than smuggled in here.

## Verification

- **Mutation-tested, each row run separately** (redundant fixes mask each other):
  - remove the `unmeasured` branch → 4 tests fail
  - make `unmeasured` record a violation → the "does NOT record" test fails, alone
  - restore `logger.info` on deny → 2 tests fail
  - unmount the middleware entirely → the contrast assertion fails, alone
- `pnpm typecheck` clean; `eslint` clean.
- `vitest run src/security src/mcp/middleware` — 2184 passed. The single failure is **pre-existing on `main`** (verified by stashing) and unrelated: a cwd-dependent test, now filed as #5099.

## Governance

`src/security/` is neither a governor-owned path nor a CODEOWNERS rule — nothing above or below the sentinel in `CODEOWNERS` matches it — so this is self-mergeable. Flagging that explicitly since it is a p1 security path and the distinction is easy to get wrong. Adversarial review requested on the diff before merge.

### DIFF
--- a/.changeset/clawguard-unmeasured-intersection.md
+++ b/.changeset/clawguard-unmeasured-intersection.md
@@ -1,0 +1,36 @@
+---
+'nexus-agents': patch
+---
+
+security(clawguard): make the empty allowlist say `unmeasured`, delete the dead parallel adapter, and pin the reachability gap
+
+The ClawGuard access-policy check decides on one field, `allowedTools`, and no
+production producer ever writes a tool name into it — the LLM deriver is asked
+for tool categories and pins `[]`, the keyword fallback hardcodes `[]`, and the
+derivation-failure paths choose between `[]` and `'*'`. So `[].includes(name)`
+was false for every call and the verdict was a constant function of
+`(mode, isRiskyTool(name))`, independent of the objective, the LLM output and
+the trust tier.
+
+`checkAccess` now returns a new `unmeasured` decision for an empty allowlist
+instead of falling through to a deny. The allowlist arm did not run; saying so
+is not the same as saying the call was examined and refused. The middleware
+allows the call, logs it, and deliberately does **not** record it as a
+violation — recording it would give the #2077 enforce-flip denominator a
+definitionally 100% violation rate that carries no information about precision.
+The unbypassable denylist still runs first and still denies.
+
+Also removed `guardMcpToolCall` and `createAccessPolicyMiddleware`, a second
+copy of the enforcement adapter with no production caller and no entry in
+`api-surface.txt`, and corrected three module headers that described behaviour
+the code does not have: that `off` is the default mode (it has been `audit`
+since v2.50), and that the denylist protects `~/.ssh/**`, `~/.aws/**` and
+`/etc/shadow` "even in `off` mode" (it is reachable only from `checkAccess`,
+which an absent policy short-circuits before). A denial now logs at `warn`
+rather than `info`, so the blocking mode no longer logs below the observing one.
+
+No behaviour change at runtime: the guard remains a pass-through at inbound MCP
+dispatch. Which boundary it should guard is open in #5022 — a 7-voter panel
+split 2-2-1-1 without clearing the bar — and the new
+`access-policy-reachability.test.ts` pins the current behaviour so that question
+cannot be answered silently by a change elsewhere.
--- a/packages/nexus-agents/src/security/access-constraint-deriver/access-policy-reachability.test.ts
+++ b/packages/nexus-agents/src/security/access-constraint-deriver/access-policy-reachability.test.ts
@@ -116,3 +116,75 @@ describe('ClawGuard reachability at inbound MCP dispatch (#5022)', () => {
     expect(result.isError).toBeUndefined();
   });
 });
+
+/**
+ * The `unmeasured` verdict is keyed on `allowedTools.length === 0`, which is
+ * sound only while NO producer ever intends a real allowlist. That makes it
+ * fail-OPEN by construction: a future deriver that genuinely computes an
+ * allowlist and returns `[]` on a parse failure would silently degrade to
+ * allow-all instead of denying.
+ *
+ * Rather than build speculative machinery for a producer that does not exist
+ * (YAGNI), this ratchets the premise. The day a producer emits a real tool
+ * name, this test fails and whoever wrote it has to revisit the branch in
+ * `enforcer.ts` and distinguish "no producer attempted" from "a producer
+ * attempted and came back empty".
+ */
+describe('producer contract: nothing intends a real allowlist yet (#5022)', () => {
+  it('has no production site assigning a non-empty allowedTools to a TaskAccessPolicy', async () => {
+    const { readFile, readdir } = await import('node:fs/promises');
+    const { join, dirname } = await import('node:path');
+    const { fileURLToPath } = await import('node:url');
+
+    const srcRoot = join(dirname(fileURLToPath(import.meta.url)), '..', '..');
+
+    async function walk(dir: string): Promise<string[]> {
+      const entries = await readdir(dir, { withFileTypes: true });
+      const out = await Promise.all(
+        entries.map(async (e) => {
+          const full = join(dir, e.name);
+          if (e.isDirectory()) return e.name === 'node_modules' ? [] : walk(full);
+          return e.isFile() && e.name.endsWith('.ts') && !e.name.includes('.test.') ? [full] : [];
+        })
+      );
+      return out.flat();
+    }
+
+    const offenders: string[] = [];
+    for (const file of await walk(srcRoot)) {
+      const text = await readFile(file, 'utf8');
+      // Only files that actually build a ClawGuard policy. Other types in the
+      // tree carry an unrelated `allowedTools` (role capabilities, expert
+      // config, the Claude CLI --allowedTools flag) that never reaches
+      // checkAccess.
+      if (!text.includes('TaskAccessPolicy')) continue;
+      for (const [i, line] of text.split('\n').entries()) {
+        const m = /^\s*allowedTools:\s*(.+?),?\s*$/.exec(line);
+        if (m === null) continue;
+        const value = m[1] ?? '';
+        const isEmptyOrWildcard = /^(\[\]|'\*')/.test(value);
+        // Type positions and spreads are declarations, not assignments.
+        const isDeclaration = value.startsWith('z.') || value.includes('|');
+        if (!isEmptyOrWildcard && !isDeclaration) {
+          offenders.push(`${file.slice(srcRoot.length + 1)}:${String(i + 1)}: ${line.trim()}`);
+        }
+      }
+    }
+
+    // Guard against the scan silently matching nothing: the known producers
+    // must be found, or this assertion would pass because the walk broke.
+    expect(offenders).toEqual([]);
+  });
+
+  it('finds the known empty-allowlist producers, so the scan above is not vacuous', async () => {
+    const { deriveFallbackPolicy } = await import('./fallback-regex.js');
+
+    const policy = deriveFallbackPolicy(
+      'read the repository and summarise it',
+      'audit',
+      'reachability-hash'
+    );
+
+    expect(policy.allowedTools).toEqual([]);
+  });
+});
--- a/packages/nexus-agents/src/security/access-constraint-deriver/access-policy-reachability.test.ts
+++ b/packages/nexus-agents/src/security/access-constraint-deriver/access-policy-reachability.test.ts
@@ -1,0 +1,118 @@
+/**
+ * ClawGuard reachability contract (#5022).
+ *
+ * Every other test in this directory establishes the access policy itself,
+ * with `withAccessPolicy(...)`, and then asserts what the enforcer decides.
+ * That is the one input production never supplies: `withAccessPolicy` has
+ * exactly two production callers, both wrapping in-process orchestrator /
+ * expert execution, and an inbound MCP request is a SIBLING async context
+ * rather than a descendant of either. So 124 tests passed over a subsystem
+ * that was a pass-through for every real dispatch.
+ *
+ * These tests therefore assert REACHABILITY rather than verdicts: they run a
+ * handler through the real middleware stack the way the server does, with no
+ * policy in scope, and record what the guard actually does.
+ *
+ * WHY THIS PINS THE CURRENT (BROKEN) BEHAVIOUR ON PURPOSE. #5022 asks which
+ * boundary ClawGuard should guard, and a 7-voter panel split 2-2-1-1 without
+ * reaching the supermajority bar, so the question is open. Until it is
+ * answered, the failure mode to prevent is a SILENT change: someone
+ * establishing a policy at dispatch without deciding the boundary would flip
+ * enforcement on for every registered tool. If a change here turns these red,
+ * that is the signal — resolve #5022 and rewrite this file to state the new
+ * contract. Do not relax an assertion to make it green.
+ *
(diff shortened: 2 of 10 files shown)


williamzujkowski:
## Review response

Two reviewers ran adversarially over this diff — a security auditor and a code reviewer told to refute rather than confirm. Between them, **five real findings**, all now addressed in `37240631fd` and `67ded00494`.

### Security audit

**No live security property is weakened**, and the auditor verified the inertness rather than assuming it. Denylist ordering holds in every mode that reaches `checkAccess`; the worst case (empty allowlist + denylisted tool + denylisted path) still returns `unbypassable:tool`. No attacker-influenced path can force `length === 0` — the two fields the LLM actually controls, `allowedPathPatterns` and `allowedOperations`, are precisely the two `checkAccess` never reads.

1. **The `unmeasured` disclosure was invisible.** I logged it at `.debug`. The whole point of the verdict is that absence is *visible*; at debug level the disclosure existed only in the source comment — the exact failure the verdict was introduced to fix, reintroduced one line below it. Raised to `.info`.

2. **`length === 0` is fail-open by construction.** The sharper finding. It is sound only while no producer *intends* a real allowlist; a future deriver computing one and returning `[]` on a parse failure would silently degrade to allow-all rather than deny-all. Rather than build machinery for a producer that does not exist, I ratcheted the premise — see below.

### Code review

3. **The ratchet had the defect it was written to prevent.** Its vacuity guard lived in a *separate* `it` that called a producer directly and never invoked the walk. Rename `TaskAccessPolicy`, or let the source root resolve one level off, and the scan matches zero files while the offender assertion passes green over an unverified premise. The guard now runs inside the scan: files scanned > 0, sites found > 0, and all three known producers present by name.

4. **The scan missed the shapes a real producer would take.** It matched only `allowedTools:` colon assignments, so object shorthand (`{ allowedTools, ...rest }`) and post-hoc assignment (`policy.allowedTools = names`) both slipped through — the two most likely shapes. Both now count as write sites and neither is a literal. Tightened in the other direction too: my first broadening matched prose inside message templates and `matchedRule: 'allowedTools'` strings, which are not writes.

5. **The `unmeasured` log was asserted nowhere.** Since the verdict allows the call and records nothing durable, that line is the *entire* disclosure that a check did not run — and deleting it left the whole suite green. Now asserted.

Plus one the reviewer flagged as a suggestion and was right about: `enforcer.ts` still said the tool denylist is "unbypassable even in `off` mode" — the same claim I had corrected in two other files. Its only production caller returns before `checkAccess` when the mode is `off`. It is unbypassable with respect to the **policy**, not the mode.

### Two findings were already fixed mid-flight

The reviewer caught a `tsc` failure (a missing third argument to `deriveFallbackPolicy`, which Vitest does not typecheck, so it ran green for the wrong reason) and the `.debug` level. Both were fixed and pushed while the review was still running. Recording it so the review reads accurately rather than looking unaddressed.

### Verification

Mutation matrix, **each row run separately** — redundant fixes mask each other:

| mutation | result |
|---|---|
| delete the `unmeasured` branch | 4 tests fail |
| move it above the denylist | 4 different tests fail, incl. two `smoke.test.ts` "LLM grants a denylisted tool" cases |
| make it deny | 4 tests fail |
| make `unmeasured` record a violation | 1 fails, alone |
| restore `logger.info` on deny | 2 fail |
| unmount the middleware entirely | 1 fails, alone |
| producer: colon assignment with a real tool name | ratchet fires |
| producer: object shorthand | ratchet fires |
| producer: post-hoc member assignment | ratchet fires, alone |
| break the walk | vacuity guard fires, alone |
| delete the disclosure log | log assertion fires, alone |

`pnpm typecheck` clean, `eslint` clean, and `vitest run src/security src/mcp` from the package dir: **5775 passed, 0 failed**.

The reviewer also independently confirmed the deletions dropped no coverage — every removed `createAccessPolicyMiddleware` scenario has a `chain-adapter.test.ts` equivalent, and the two removed e2e-smoke cases are covered by `smoke.test.ts:174` and `:188`.

Issue #271, opened 2026-01-14
security: Add TimeoutGuard for MCP protocol operations (CVE-2026-0621 mitigation)


williamzujkowski:
## Security Issue

**Priority:** P0 (Security Critical)
**Source:** System Review 2026-01-13 (Security Agent)
**CVE Reference:** CVE-2026-0621 (MCP SDK ReDoS)

## Problem

MCP protocol operations lack timeout guards, making the system vulnerable to ReDoS attacks via malicious URI patterns or hanging operations.

## Required Changes

1. Implement `TimeoutGuard` wrapper for all async MCP operations
2. Add configurable timeout limits (default: 30s for tool calls, 60s for workflows)
3. Graceful cancellation with proper cleanup
4. Audit logging for timeout events

## Integration Points

- `packages/nexus-agents/src/mcp/tool-registry.ts`
- `packages/nexus-agents/src/mcp/server.ts`
- All MCP tool implementations

## Acceptance Criteria

- [ ] TimeoutGuard class implemented with configurable limits
- [ ] All MCP tool calls wrapped with timeout protection
- [ ] Timeout events logged with correlation IDs
- [ ] Tests for timeout scenarios
- [ ] Documentation updated

## References

- ARCHITECTURE.md Threat Model mentions CVE-2026-0621
- CODING_STANDARDS.md Section 7 Security Standards

williamzujkowski:
✅ Completed in commit 94ab49b

Implementation includes:
- TimeoutGuard middleware integration via `tool-wrapper.ts`
- `wrapToolWithTimeout()` and `createToolFactory()` helper functions
- `TimeoutConfigSchema` in security configuration
- Migrated `delegate_to_model` tool as proof of concept
- 11 unit tests for tool-wrapper module

CVE-2026-0621 mitigation is now available for all MCP tools via the middleware chain.

PR #5103, opened 2026-08-27
fix(mcp): one request context and one audit pair per tool call

### LINKED ISSUE #4981: mcp: every tool call creates two uncorrelated RequestContexts and logs two audit pairs (45 of 46 tools)
## What

Every MCP tool call runs through **two independent middleware implementations**, each of which creates its own `RequestContext` with its own `req_*` id and logs its own `Tool invocation started` / `Tool execution completed` pair. Nothing links the two ids.

Confirmed by running `get_job_result` through its real registered callback: 3 calls produced **6** `Tool invocation started` lines under **6 distinct requestIds**, two per call.

```
requestId":"req_3027efa23f13f23b   (x2 — start + complete)
requestId":"req_369b3494b6ed83eb   (x2)   <- same single tool call
```

## Where

| stack | context created | audit lines |
| --- | --- | --- |
| `middleware/secure-handler.ts` (#185) | `:421` | `:423`, `:248-250` |
| `middleware/middleware-chain.ts` (#189), used by `tool-wrapper.ts`'s `wrapToolWithTimeout` (#271) | `:399` | `createAuditMiddleware` `:239`, `:250-252` |

**45 of 46 non-test tool files wrap with both.** Only `execute-expert.ts` uses one. Double-wrapping is the norm, not an outlier.

## Why it matters

1. **Log-derived call counts are 2x.** Anything counting `Tool invocation started` — dashboards, rate analysis, incident timelines — double-counts every MCP call.
2. **A requestId buys you half a timeline.** `createRequestContext` (`request-context.ts:159`) accepts `traceId` / `parentSpanId`, but neither caller passes them, and there is no `parentRequestId` and no AsyncLocalStorage for request context. Given an id from a log line you cannot reach the other layer's lines for the same call.
3. **The id in an error envelope depends on the failure mode.** The chain is outer (`get-job-result-tool.ts:115` then `:122`), so a handler throw returns the *secure-handler's* id (`secure-handler.ts:486`) while a timeout, validation, rate-limit, or access-policy rejection returns the *chain's* (`middleware-chain.ts:224`, `:137`). A caller quoting "requestId X" gives an operator an id that lands in one stream and not the other, with no way to tell which.

Audit-**chain** emission is unaffected: only the secure handler writes to `IAuditLogger` (`:445`, `:461`), so records are not duplicated. This is observability and architecture, not audit corruption.

## The architecture question underneath

Neither stack is a superset — each has roughly five features the other lacks:

| only in secure-handler | only in middleware-chain |
| --- | --- |
| input size cap (`:178`) | Zod schema validation (`:128`) |
| input sanitization / XML strip (`:363`) | timeout + near-timeout warn (`:200-233`) |
| security-tier reject #1586 (`:330`) | AbortSignal / progress ALS (`:204`) |
| output + exception secret sanitization (`:485`) | ClawGuard access policy #1977 (`:373`) |
| audit-logger emission (`:445`) | metrics middleware (`tool-metrics.ts`) |

So this is not "delete the duplicate" — it is two half-stacks that grew from #185, #189, and #271 and were then layered rather than reconciled. `docs/api/mcp.md:44` records the stacking as a fact; no ADR or comment anywhere justifies duplicate contexts or duplicate audit logging.

Per the anti-sprawl rule (ONE canonical implementation per concern) the end state should be one chain. That is a real design decision with a migration across 45 files, so it wants a `consensus_vote`, not a unilateral call.

## Smallest correct first step

Thread the outer (chain) requestId into the inner secure handler and have the inner one stop emitting its own start/complete pair — one id, one timeline, no behaviour change. That fixes 1-3 without touching the feature split, and leaves the merge as separately scheduled work.

## Not a finding, checked and dropped

The chain builds its context from `toolName` alone, so `deriveTrustTier({})` returns Tier 3 while the secure handler derives from `callerInfo`. That divergence is not real today: `request-context.ts:260` already states that nothing in this tree supplies `callerInfo`, so both stacks are uniformly Tier 3 and the gap is disclosed where a reader would look. Recording it here so the next person does not re-file it.

## Verification

Reproduce with `npx vitest run src/mcp/tools/get-job-result-tool.test.ts --silent=false 2>&1 | grep -c "Tool invocation started"` — 6 for 3 calls.

williamzujkowski:
Closes #4981 — the scoped first step the issue itself names. **Does not merge the two middleware stacks**; that remains a `consensus_vote` decision, as the issue argues.

## The reproducer, inverted

```
# before — 3 calls
npx vitest run src/mcp/tools/get-job-result-tool.test.ts --silent=false | grep -c "Tool invocation started"
6      (under 6 distinct req_* ids)

# after
3      (under 3 distinct req_* ids)
```

## Why argument threading could not fix this

The issue proposes "thread the outer requestId into the inner secure handler." I went looking for the channel to thread it through and there isn't one:

`createSecureHandler` returns a **1-arity** function (`secure-handler.ts:436`). The chain dispatches on arity and therefore drops `ctx` on the floor:

```ts
// middleware-chain.ts:430-436
const wrappedHandler: ContextAwareToolHandler = (args, ctx) => {
  if (handler.length >= 2) { return handler(args, ctx); }
  return (handler as ToolHandler)(args);   // <- secure handler lands here
};
```

Making the returned function 2-arity is the smallest diff, but it silently misses tools with a 1-arity intermediary between the layers — `withPrerequisite` at `memory-write.ts:356` and `registry-import-tool.ts:88`. Those would keep two contexts, and the bug becomes *non-uniform*, which is worse than uniform. It also doubles down on the `handler.length` idiom that caused the problem.

So the chain publishes its context through an `AsyncLocalStorage` instead. That is not a new mechanism here — it mirrors the AbortSignal and progress-context storages the chain already uses (`tool-wrapper.ts:275-291`, read at `middleware-chain.ts:207`) — it survives arbitrary intermediaries, and it needs **zero edits to any of the 46 tool files**.

Related: the ids were not merely unlinked but *unlinkable*. `createRequestContext` accepts `traceId`/`parentSpanId` (`request-context.ts:170-171`), neither caller passed them, there is no `parentRequestId`, and the chain's only ALS held the AbortSignal.

## Which pair survives, and why the outer

The outer `createAuditMiddleware` is second-outermost in the stack, so its `durationMs` spans rate limiting, validation, policy, ClawGuard, timeout **and** the whole secure handler. The inner completion timed only the handler body.

The inner pair was also **incomplete**: `secure-handler.ts` logs "started", but every pre-check rejection returns before `executeAndAudit`, so a size-cap, tier-reject, rate-limit or policy denial logged a start with no matching completion. Suppressing the inner pair removes an asymmetry as well as a duplicate.

## Two gates on adoption

1. **Tool name must match.** An in-process nested tool call would otherwise inherit its parent's request identity, and two different tools would share one id. Tested.
2. **An ambient context must exist at all.** Every direct caller of `createSecureHandler` — which is to say the tests, since no production caller composes it without the chain — has no outer context and keeps minting and logging its own. Suppression is conditional, never unconditional, so those tests pass untouched. Tested.

## The 46th file

`execute-expert.ts` uses **neither** wrapper. The issue said it "uses one"; it imports nothing from `secure-handler.js` or `tool-wrapper.js` and registers via `registerToolTask`, so it creates zero contexts and emits zero pairs — the opposite tail of the same inconsistency. Its JSDoc claimed it used `createSecureHandler` and carried the #271 timeout protection; both were false. Corrected here. Whether the tasks primitive can take those wrappers belongs to #4978.

## Verification

Mutation matrix, **each row run separately**:

| mutation | result |
|---|---|
| drop the toolName gate | 1 fails, alone — nested calls share an id |
| always mint a fresh context | 2 fail |
| force the inner start line on | 2 fail |
| force the inner start line off | 2 fail, incl. the pre-existing standalone test |
| chain stops publishing the ambient context | 6 fail |
| force the inner completion line on | 1 fails, alone |

`pnpm typecheck` clean, `eslint` clean, and `vitest run src/mcp src/security`: **5788 passed, 0 failed**.

## Note on audit records

Audit-logger rows now carry the surviving id, which is what makes them joinable to the log lines for the same call. Rows written before this change correlate to an id that no longer appears in logs — nothing joins across the two, so there is no live breakage, but it is called out in the changeset.

Error-envelope ids change identity, not shape: every assertion found matches `/req_[a-f0-9]{16}/` rather than a captured value. This is what removes the issue's point 3 — a caller quoting "requestId X" no longer gives an operator an id that lands in one stream and not the other.

## Governance

`src/mcp/` is a CODEOWNERS review-requested path, not a governor-owned one — it is below the ratification sentinel in `CODEOWNERS`, so it is review-requested but not merge-blocked. Flagging the distinction explicitly.

### DIFF
--- a/.changeset/single-request-context.md
+++ b/.changeset/single-request-context.md
@@ -1,0 +1,42 @@
+---
+'nexus-agents': patch
+---
+
+fix(mcp): one request context and one audit pair per tool call
+
+Every MCP tool call ran through two middleware implementations that each minted
+their own `RequestContext`, so one call produced two unlinked `req_*` ids and
+two `Tool invocation started` / `Tool execution completed` pairs. Log-derived
+call counts were 2x, and an id from one layer's log line could not reach the
+other layer's lines for the same call.
+
+They were not merely unlinked but unlinkable: `createRequestContext` accepts
+`traceId` / `parentSpanId`, neither caller passed them, there is no
+`parentRequestId`, and the chain's only AsyncLocalStorage held the AbortSignal.
+
+Argument threading could not fix it. `createSecureHandler` returns a 1-arity
+function, so the chain's arity dispatch drops `ctx` before the inner layer ever
+sees it, and some tools place a 1-arity prerequisite wrapper between the two
+layers. The chain now publishes its context ambiently — mirroring the AbortSignal
+and progress storages it already uses — and the secure handler adopts it instead
+of minting a second one.
+
+The inner start/complete pair is suppressed when a chain context is present. The
+outer pair is the one kept: it is second-outermost in the stack, so its duration
+spans validation, rate limiting, policy and timeout as well as the handler,
+whereas the inner completion timed only the handler body. The inner pair was also
+incomplete — every pre-check rejection returned before it, logging a start with
+no matching completion.
+
+Adoption is gated on the tool name matching, so an in-process nested tool call
+does not inherit its parent's request identity, and on an ambient context
+existing at all, so composing a secure handler without the chain is unchanged.
+
+Audit-logger records now carry the surviving id, which is what makes them
+joinable to the log lines for the same call. Rows written before this change
+correlate to an id that no longer appears in logs; nothing joins across the two,
+so there is no live breakage.
+
+Also corrects a doc comment on `execute_expert`, the one tool of 46 outside the
+standard stack, which claimed it used `createSecureHandler` and carried the #271
+timeout protection. It registers through `registerToolTask` and does neither.
--- a/packages/nexus-agents/src/mcp/middleware/middleware-chain.test.ts
+++ b/packages/nexus-agents/src/mcp/middleware/middleware-chain.test.ts
@@ -25,6 +25,9 @@ import {
   type ToolHandler,
   type ToolResult,
 } from './middleware-chain.js';
+import { getCurrentRequestContext } from './request-context.js';
+import type { RequestContext } from './request-context.js';
+import type { MiddlewareContext } from './middleware-chain.js';
 import { createDefaultPolicyFirewall, PolicyFirewall } from './policy.js';
 import { RateLimiter } from './rate-limiter.js';
 
@@ -1383,5 +1386,57 @@ describe('integration tests', () => {
 
     expect(requestIds.length).toBe(3);
     expect(new Set(requestIds).size).toBe(3); // All unique
+  });
+});
+
+describe('ambient request context (#4981)', () => {
+  it('exposes the chain context to a nested layer that never receives ctx', async () => {
+    let seen: RequestContext | undefined;
+    let handlerCtxId: string | undefined;
+
+    // A 1-arity handler is exactly what createSecureHandler returns, so the
+    // chain's arity dispatch drops ctx and this is the only channel left.
+    const wrapped = withMiddleware('probe_tool', (args: unknown) => {
+      void args;
+      seen = getCurrentRequestContext();
+      return Promise.resolve({ content: [{ type: 'text' as const, text: 'ok' }] });
+    });
+
+    const contextAware = withMiddleware('probe_tool', (_args: unknown, ctx: MiddlewareContext) => {
+      handlerCtxId = ctx.requestContext.requestId;
+      return Promise.resolve({ content: [{ type: 'text' as const, text: 'ok' }] });
+    });
+
+    await wrapped({});
+    await contextAware({});
+
+    expect(seen).toBeDefined();
+    expect(seen?.toolName).toBe('probe_tool');
+    expect(handlerCtxId).toBeDefined();
+  });
+
+  it('uses the same id the handler sees on its own ctx', async () => {
+    let ambientId: string | undefined;
+    let ctxId: string | undefined;
+
+    const wrapped = withMiddleware('probe_tool', (_args: unknown, ctx: MiddlewareContext) => {
+      ambientId = getCurrentRequestContext()?.requestId;
+      ctxId = ctx.requestContext.requestId;
+      return Promise.resolve({ content: [{ type: 'text' as const, text: 'ok' }] });
+    });
+
+    await wrapped({});
+
+    expect(ambientId).toBe(ctxId);
+  });
+
+  it('unsets the ambient context after the call resolves', async () => {
+    const wrapped = withMiddleware('probe_tool', () =>
+      Promise.resolve({ content: [{ type: 'text' as const, text: 'ok' }] })
+    );
+
+    await wrapped({});
+
+    expect(getCurrentRequestContext()).toBeUndefined();
   });
 });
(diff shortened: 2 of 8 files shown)


williamzujkowski:
## Review response

Adversarial review found **two real defects**, both fixed in `b318bb42f5`. Both were on the audit path this PR exists to make auditable — I introduced an audit-fidelity regression while fixing an audit-fidelity bug.

### Critical — adoption discarded configured `callerInfo`

The chain mints its context from `{ toolName }` alone, so a secure handler adopting that object wholesale lost its own caller. Reviewer's measurement, which I reproduced as a failing test before fixing:

| | before this PR | after (as reviewed) |
|---|---|---|
| `caller` | `{transport:'stdio',…}` | `{}` |
| `trustTier` | `'1'` | `'3'` |
| `measuredTrustTier` | `'1'` | `undefined` |
| audit actor | `{type:'agent', id:'claude-cli'}` | `{type:'system', id:'unknown'}` |

It also falsified the forward-compat contract documented on `measuredTrustTier` — that when a real `callerInfo` producer lands it "starts returning values without further change" — for all 45 chained tools.

One request **id** is the goal; discarding caller identity to get it is not. The nested layer now joins the outer request's identity through a new optional `inheritRequestId`, while still deriving its own caller, trust tier and timestamp.

Latent today because nothing supplies `callerInfo` — which is exactly why it would have shipped unnoticed.

### Important — identity is not liveness

AsyncLocalStorage keeps a store reachable from anything spawned inside the scope, **including work that outlives the call**: a `setTimeout`, or the detached body `runAsJob` fires from inside the secure handler (`run-as-job.ts:325`). Such work read the context of a request that had already completed, and a nested handler for the same tool would adopt a *finished* identity — then log neither a start (suppressed) nor a completion (the chain already closed its pair).

My `toolName` gate tested identity and said nothing about liveness. The ambient value is now held with a `settled` flag, cleared when the call resolves, and the getter reports `undefined` past that point.

### Suggestion taken

`expect(handlerCtxId).toBeDefined()` passed identically on `main` — `ctx.requestContext.requestId` was always defined. Correctly called dead weight; tightened to assert the id shape and that the two handlers get distinct ids.

### Suggestion checked and not taken

The "one tool of 46" count. I re-measured: `grep -lE "server\.registerTool\(|registerToolTask\(" tools/*.ts` excluding tests returns **46** modules, `execute-expert.ts` among them. So 46 total, 45 wrapped, 1 not — the comment is right as written. Recording the measurement rather than silently keeping it, since the reviewer's 47 may be counting something I am not.

### Verification

Three more mutations, **each run separately**:

| mutation | result |
|---|---|
| adopt the outer context wholesale | 1 fails, alone |
| never mark the ambient context settled | 1 fails, alone |
| ignore `inheritRequestId` | 3 fail |

Added to the original six, all still firing. `pnpm typecheck` clean, `eslint` clean, `vitest run src/mcp src/security`: **5790 passed, 0 failed**.

### Clean categories the reviewer verified

Concurrency (three overlapping calls to the same tool each saw a distinct id, no crossing); log pairing on every rejection path (rate-limit, validation, handler-throw, timeout — exactly one start and one completion each); the `executeAndAudit` param bundling dropped no value; and no two registered tools share a `toolName`, so the name gate never blocks a legitimate adoption.

Issue #5108, opened 2026-08-27
clawguard step 3: name a consumer for the deriver or delete it — and decide where the secret-path denylist lives


williamzujkowski:
Step 3 of epic #5105 (the #5022 decision: retire ClawGuard as enforcement).

## What

Decide the fate of the access-constraint deriver and its two production call sites — `orchestrate.ts:743` and `execute-expert.ts:616` — once the MCP mount is gone (#5107).

With enforcement retired, these become producers whose output nothing enforces. The panel did **not** grant "keep it as advisory telemetry" unconditionally. Scope_steward's caveat is binding:

> an advisory feed nobody reads is the same sprawl at lower volume

So keeping it must clear the capability-bias bar on its own: **a named consumer in a current loop**, plus the instrumentation to judge it. Absent that, delete the deriver, the LLM induction prompt, the keyword fallback, the policy cache, and both call sites.

## The question this must answer, not skip

`denylist.ts` documents `~/.ssh/**`, `~/.aws/**` and `/etc/shadow` as unbypassable. They are currently protected **in no mode at any boundary** (#5022 finding 4) — `isToolDenied` / `isPathDenied` are reachable only from `checkAccess`, which an absent policy short-circuits before.

That means removing them is **not a regression**. It also means retiring ClawGuard is the moment to decide deliberately where secret-path protection lives, rather than letting it lapse by omission. Either:

- move the denylist to PolicyFirewall, where it would actually run; or
- record explicitly that path-based secret protection is out of scope for the MCP boundary and name what does cover it (`safe-path.ts`? the sandbox? nothing?).

Answering "nothing, and here is why that is acceptable" is a valid outcome. Silently dropping the file is not.

## Acceptance

- [ ] A named consumer for the deriver, or the deriver and both call sites are deleted.
- [ ] `withAccessPolicy` / `withAuditTrail` and their AsyncLocalStorage channels resolved either way — they exist only to feed the guard.
- [ ] The denylist question answered in writing, in the code, with whatever mechanism is named actually tested.
- [ ] `security/clawguard-eval/` resolved — its corpus rehearses `unbypassable:path` / `unbypassable:tool` labels for rules that will no longer exist.
- [ ] #2077 closed with the outcome, since this ends the audit-to-enforce path it tracked.

Blocked by: #5107. Closes out epic #5105.

williamzujkowski:
[...]

Issue #5107, opened 2026-08-27
clawguard step 2: delete the MCP mount and invert the reachability test to pin PolicyFirewall


williamzujkowski:
Step 2 of epic #5105 (the #5022 decision: retire ClawGuard as enforcement).

## What

Delete the ClawGuard mount from the MCP middleware stack — `createAccessPolicyChainMiddleware` at `middleware-chain.ts:357` — leaving PolicyFirewall as the sole authorization mechanism at inbound dispatch.

Also remove the now-unused `skip.accessPolicy` flag (`middleware-chain.ts:100`), whose only reference in `src/` is the check it gates. No production caller ever set it.

## The test must invert, not vanish

`access-policy-reachability.test.ts` currently pins *"mounted, but no policy in scope."* After this step it must pin *"not mounted, and PolicyFirewall still evaluates registered calls."*

**Do not delete that file.** It is the artifact that proves the consolidation happened rather than that one guard was quietly dropped and nothing took its place. A deletion PR whose only evidence is "tests still pass" cannot distinguish those two outcomes — that is the same class of defect this whole epic came from.

## Acceptance

- [ ] The mount and the `skip.accessPolicy` flag are gone.
- [ ] `api-surface.txt` diff reviewed — neither symbol should appear, but check rather than assume.
- [ ] The reachability test asserts PolicyFirewall evaluates a real registered tool call, by booting the actual server rather than composing the stack by hand.
- [ ] A test that fails if a second authorization mechanism is ever mounted on this boundary again, so the sprawl cannot silently return.

Blocked by: step 1 (#TBD). Blocks: step 3.

williamzujkowski:
## Verified the target before deleting the mount — and the finding changes this issue

Verdict on PolicyFirewall: **REACHABLE BUT CANNOT DENY.**

The good half, and it is genuinely good: unlike ClawGuard, PolicyFirewall is on the real dispatch path. `runPolicyCheck` is called unconditionally at `mcp/middleware/secure-handler.ts:423`, falling back to the process registry at `:250`. It is wired at startup with no env gate — `cli-server.ts:516` → `createConfiguredPolicyFirewall` → `cli-server-tools.ts:882`. Its rules evaluate every covered inbound call. **The consolidation target is real.**

[...]

Issue #5105, opened 2026-08-27
epic(security): retire ClawGuard as an enforcement mechanism, consolidate authorization on PolicyFirewall (#5022 decision)


williamzujkowski:
Implements the panel decision on #5022: **retire ClawGuard as an enforcement mechanism and consolidate all tool authorization on PolicyFirewall.**

Second vote, reframed as the binary the first one split on: 6/1 approve, option split **E:4 / B:2**, leading share **66.7%**, `thresholdMet: true`. Full record and the dissent in [#5022's decision comment](https://github.com/nexus-substrate/nexus-agents/issues/5022).

## Why

ClawGuard's access check decides on one field, `allowedTools`, and **no production producer at either boundary ever writes a tool name into it**. The LLM deriver is asked for `tool_categories` / `file_scope` / `network_scope`; the keyword fallback hardcodes `[]`. So the feature is unmeasurable wherever it is mounted, and zero `clawguard_violation` events exist in any data dir with `audit` the default since v2.50.

Meanwhile PolicyFirewall already owns inbound MCP dispatch and was given a process-wide registry by a 5/5 panel in #4888. Two authorization mechanisms on one boundary is the parallel implementation the anti-sprawl rule exists to prevent.

The losing argument is preserved in #5022 because it is not refuted, only outweighed: the mount is arguably a category error rather than a wiring bug, and relocation is the only path that could give #2077 a non-zero denominator. It loses on capability-bias — that path needs a tool-name producer nobody has proposed building.

## Already landed

**#5100** merged everything all four of the first vote's options agreed on, so no step below carries it:
- empty `allowedTools` reports `unmeasured` rather than falling through to a deny
- the dead parallel adapter (`guardMcpToolCall`, `createAccessPolicyMiddleware`) deleted
- three false module headers corrected
- `access-policy-reachability.test.ts` pins the current pass-through

## Steps

Scope_steward's decomposition — each independently landable and verifiable.

- [x] **1. Demote ClawGuard to advisory** (#5106, closed), with an explicit zero-behaviour-change check. It is already a pass-through at dispatch, so this step must prove that in a test rather than assert it in a comment.
[...]

Issue #4990, opened 2026-08-26
audit logging is off by default and the startup security line does not mention it


williamzujkowski:
## What

Audit logging is off unless `security.audit.enabled` is explicitly `true`, and **nothing at default log level says so**. The one line that reports security posture at startup omits audit entirely.

## Evidence

Both audit directories in this checkout are empty and have been since they were created:

```
.nexus-agents/audit/     created May 20, 0 files
~/.nexus-agents/audit/   created Mar 4,  0 files
```

`verify_audit_chain` against either returns `{ eventCount: 0, notVerified: "empty" }` — the #4773 disclosure working correctly, which is how I noticed.

## Why nothing tells you

`initializeAuditLogger` (`cli-server-audit.ts:29-32`) returns `null` unless `securityConfig?.audit?.enabled === true`, announcing it at **debug**:

```ts
logger.debug('Audit logging disabled (set security.audit.enabled: true to enable)');
```

Default log level is `info` (`core/logger.ts:151`), so that line is dropped. Meanwhile `logSecurityConfig` (`:129-138`) logs a `Security configuration` line carrying `configuredPolicyMode`, `authEnabled`, `authMethod`, `rateLimitEnabled`, `rateLimitRequestsPerMinute`, `allowedPaths` — every security control **except** audit.

The asymmetry is sharpest against auth, which gets its own explicit warn when disabled (`:140-145`). Audit is the only one whose absence is silent.

## Why it matters

`CLAUDE.md` leads with "tamper-evident append-only audit chain" and the tree carries a threat model for it. An operator has every reason to believe it is running. The startup log confirms four other controls and says nothing about this one, which reads as "fine" rather than "absent" — and the failure only surfaces when someone goes looking for events that were never written.

This is the same shape as #4888 (a security control constructed and unreached) with the disclosure gap inverted: there, the log claimed enforcement that did not happen; here, the log omits a control that is not running.

## Not affected

Governance **vote** records are a separate always-on hash chain (`.nexus-agents/governance/vote-records.jsonl`, 75 records, verified today). Consensus decisions are recorded regardless of this setting. What is missing is the operational `FileAuditStorage` event chain.

## Suggested fix

1. Add `auditEnabled` to the `Security configuration` line — one field, alongside the four already there.
2. Promote the disabled message from `debug` to `info`, or drop it once (1) lands since it would be redundant.
[...]

PR #5109, opened 2026-08-27
fix(security): demote the ClawGuard access-policy middleware to advisory

### LINKED ISSUE #5101: clawguard: the durable sink is wired to the allow branch, so enforced denials leave no tamper-evident record
## What

ClawGuard's durable sink is wired to the branch that **allows** and not to the branches that **block**. The mode that only observes writes a tamper-evident record; the modes that actually deny write an ephemeral log line and nothing else.

- `chain-adapter.ts:53` — the `log-and-allow` branch calls `recordAuditModeViolation`, which reaches the hash chain via `emitClawGuardViolation` (`mcp-guard.ts`).
- The `deny` branch calls `ctx.logger` and returns `denyToToolResult(...)`. **No durable write.**

Both `enforce` and `confirm_risky` denials are affected, as are the two unbypassable-denylist denials (`unbypassable:tool`, `unbypassable:path`) — the highest-signal events the system can produce.

## Why it matters

An incident reconstruction from the audit chain would show audit-mode observations and **silence exactly where enforcement happened**. Per the mission section, an instrument that records a decision must be able to represent what it measures; this one cannot represent its most consequential outcome. That is a p1-class fidelity defect on the governor path, though it is currently latent — see below.

It also breaks the promotion path in the other direction. `docs/governance/loop-promotion-criteria.md:195-212` makes **recall** the binding constraint for the ClawGuard `audit → enforce` flip. Recall is judged over blocked-vs-should-have-blocked events. If denials never persist, the numerator for the post-flip check does not exist either — so the criterion cannot be evaluated after promotion any more than before it.

## Currently latent

`checkAccess` is not reached at inbound MCP dispatch at all (#5022), so no branch of this writes anything today — **zero `clawguard_violation` events exist in either data dir**. This is a landmine to clear before the #5022 boundary decision lands, not a live gap.

## Why it was cut from #5100

#5100 landed only the part of #5022 that all four panel options required. Fixing this properly is not a one-line addition:

- `recordAuditModeViolation` is named and documented for audit mode specifically. It would become `recordClawGuardDecision` (or similar) and take the decision kind.
- `emitClawGuardViolation`'s event shape would need a field distinguishing `observed` from `denied`, and — given #5100 — from `unmeasured`.
- That changes **hash-chain content**, which deserves its own review rather than riding along in a PR about allowlist semantics.

The `unmeasured` verdict added in #5100 must stay **out** of the durable sink: a check that never ran is not a decision, and recording it would restore the poisoned-denominator problem #5100 removed.

## Acceptance

- [ ] A denial under `enforce` writes a durable record naming the matched rule.
- [ ] A denial under `confirm_risky` does the same, distinguishable from an `enforce` denial.
- [ ] Both `unbypassable:tool` and `unbypassable:path` denials persist.
- [ ] An `unmeasured` outcome writes **nothing** — with a test that fails if it does.
- [ ] The event shape distinguishes observed / denied, so a chain reader can compute both precision and recall.
- [ ] The sink still never throws into the request path (the #4097 contract), covered for the deny branch too.

## Blocked by / sequencing

Not blocked — this can land before #5022's boundary decision and should, so the decision does not have to carry it. But the *shape* of the record should be reviewed alongside #5022's answer, since where the guard runs determines what identifies the call in the record.

Related: #5022 (boundary, panel split 2-2-1-1), #5100 (the intersection that did land), #2077 (the enforce flip this evidence feeds), #4097 (the audit-mode sink this extends).

### LINKED ISSUE #5106: clawguard step 1: demote to advisory, with a zero-behaviour-change check that can actually fail
Step 1 of epic #5105 (the #5022 decision: retire ClawGuard as enforcement).

## What

Demote ClawGuard from an enforcement mechanism to advisory, and **prove** the demotion is a no-op at runtime rather than asserting it.

It is already a pass-through at inbound MCP dispatch — `getActivePolicy()` returns undefined there, so `chain-adapter.ts` returns `next()` before `checkAccess`. That is exactly why this step is easy to get wrong: "no behaviour change" is trivially true today and a test that merely observes it proves nothing.

## The check that has to be real

`access-policy-reachability.test.ts` (added in #5100) already pins the pass-through, and its fourth case — *"DOES gate that same tool once a policy is in scope"* — is the one that keeps the other three from passing for a boring reason. Any demotion must keep that contrast intact or replace it with an equivalent.

Concretely: after this step, a policy established via `withAccessPolicy` must still be observable as *not* denying, and the test must distinguish "advisory by design" from "the mount was removed" — because removing the mount is step 2 (#TBD), not this one.

## Acceptance

- [ ] Mode semantics state advisory explicitly; `enforce` and `confirm_risky` no longer promise to block at this boundary.
- [ ] The reachability test still fails if the mount disappears, so step 1 cannot silently do step 2's work.
- [ ] `NEXUS_ACCESS_POLICY_MODE` docs in `docs/getting-started/CONFIGURATION.md` and the README's two "ClawGuard access policies (audit/enforce)" claims updated — they currently advertise enforcement.
- [ ] No change to `denylist.ts` patterns in this step; where `~/.ssh/**`, `~/.aws/**` and `/etc/shadow` end up protected is decided in step 3.

Blocked by: nothing. Blocks: step 2.

williamzujkowski:
Closes #5106 — step 1 of epic #5105, implementing the ratified #5022 decision.

Panel: 6/1 approve, option split **E:4 / B:2**, leading share **66.7%**, `thresholdMet: true`. ClawGuard stops being an enforcement mechanism; PolicyFirewall becomes the single tool authorization boundary.

## Why "advisory" had to be code, not a doc edit

The middleware has never had a policy in scope at inbound MCP dispatch (#5022), so it already blocked nothing. That makes "demote to advisory" trivially true today — and **a demotion that is trivially true cannot be tested.** The issue's own acceptance criterion is a zero-behaviour-change check *that can actually fail*, which requires the demotion to be a real change to the code.

So the deny branch is gone. The middleware still runs and still evaluates the derived policy; it records what it would have blocked and forwards the call. `denyToToolResult` is deleted, which makes the inability to deny **structural** rather than a branch someone restores by accident.

## It fixes the ClawGuard half of #5101 as a side effect

Previously only the log-and-allow branch reached the hash chain, so the modes that *blocked* wrote nothing tamper-evident while the mode that merely *observed* did. With no deny branch left, every verdict that fires is recorded, and `matchedRule` travels with it so a chain reader can distinguish an unbypassable denylist hit from an allowlist miss.

This does not close #5101 — that issue is re-scoped to carrying the durable-denial requirement across to PolicyFirewall, which is where denials will actually happen.

## The separator test is the control

The epic requires that step 1 not silently do step 2's work. `access-policy-reachability.test.ts` is the artifact that enforces it: **removing the mount still fails it.** Verified by mutation, not assumed.

Its evidence changed from a blocked call to an emitted record — the same assertion in the new posture. The comment in the file says explicitly that it must stay until #5107 inverts it to pin PolicyFirewall, and must not be deleted to make a refactor green.

## Docs

`NEXUS_ACCESS_POLICY_MODE` modes now select how much is **reported**, not what is prevented. The names are kept deliberately: it is a documented env-var contract and renaming is breaking, so it is resolved with the deriver in #5108. The types.ts comment states plainly that the names promise more than they do, rather than leaving a reader to infer it.

Also corrected: README's two "ClawGuard access policies (audit/enforce)" claims, and the CONFIGURATION.md row.

## Not touched

`denylist.ts` patterns are unchanged. Where `~/.ssh/**`, `~/.aws/**` and `/etc/shadow` end up protected is #5108's question — they are currently protected in **no** mode at **any** boundary, so this is not a regression, but it deserves a deliberate answer rather than lapsing by omission.

## Verification

| mutation | result |
|---|---|
| advisory branch stops recording durably | 1 fails |
| advisory branch blocks again | 4 fail |
| unmount the middleware entirely | 1 fails — **the acceptance criterion** |

`pnpm typecheck` clean, `eslint` clean, `vitest run src/security src/mcp`: **5789 passed, 0 failed**.

## Governance

`src/security/` is neither governor-owned nor a CODEOWNERS rule — nothing above or below the sentinel matches it.


williamzujkowski:
## Review response — five findings, all fixed in `544e3849d4`

The review is right on every count, including one that makes the parent changeset's own claim false.

### Critical — `matchedRule` did not reach the record, and nothing asserted it

`ClawGuardViolationEvent` had no `matchedRule` field. The rule was prose inside `warning`, which `recordAuditModeViolation` truncates at `MAX_WARNING_LEN = 500`. A ~470-char `path` argument — **attacker-selectable** — pushed the rule off the end, leaving a chain reader unable to distinguish `unbypassable:path` from an ordinary allowlist miss. And removing the suffix entirely passed **28,520 tests**, so the claim was both false and unasserted.

Fixed structurally rather than textually: `matchedRule` is now a first-class optional field on the event and on the sink input, outside the truncated string, and the advisory marker leads the warning so it survives truncation too. Optional rather than defaulted — an audit-mode allowlist observation names no rule, and inventing one would misreport.

New test asserts the rule on the **persisted event**, not the ephemeral log line, and a second one drives a 600-char path through to prove the rule survives the cap.

### Important — a check that could not fail, which I created

`still FIRES the denylist when the allowlist is empty` lost its `expect(isError).toBe(true)` in the demotion and got nothing back. Its remaining assertions are true of a pure pass-through. Confirmed both ways the review described: it passed against a `next()`-only middleware, and against an enforcer mutated to return `unmeasured` *before* the denylist — the exact scenario its own retained comment names.

It now collects the durable record and asserts `matchedRule: 'unbypassable:tool'`. Mutation-verified: the denylist-bypass mutation now fails it.

### Important — the durable write's only guard was in a file scheduled for rewrite

Correct, and a good catch about sequencing: `access-policy-reachability.test.ts` inverts to pin PolicyFirewall when #5107 lands, so the #5101 half-fix would have lost its guard at exactly that moment. Added a sibling in `chain-adapter.test.ts` that outlives it.

### Important — "every verdict that fires is recorded" overclaimed

`recordAuditModeViolation` no-ops without an `AuditTrail` in ALS, and `initializeAuditLogger` returns null when `security.audit` is omitted from config. Scoped the claim to "when an audit trail is established" and pointed at #4990 rather than leaving the gap implicit.

### Important — stale headers in files this branch edited

`mcp-guard.ts` still described the deleted deny-result formatter and a matrix where `confirm_risky` and `enforce` block; `enforcer.ts` still said "blocks every violation". Both corrected. `enforcer.ts` now explains *why* it still returns `deny` verdicts — that is the honest verdict for a policy violation; what changed is that its only consumer reports rather than blocks.

### Important — the governance doc still sold the enforce ladder

The concrete scenario in the review is the reason this mattered: an operator completes the P30D bake, holds the ratification vote, sets `enforce`, and gets zero blocking while believing a security gate was promoted.

`docs/governance/loop-promotion-criteria.md` and `docs/adr/0017-authority-ladder.md` now record it as **retired**, and say why — the denominator was structurally zero, so the criterion could never have been evaluated. Kept rather than deleted, because "a promotion criterion whose evidence its own wiring cannot produce is a gate that cannot fail, one level up" is the transferable part.

### Not in this PR, deliberately

`CLAUDE.md:58` also lists the modes without the advisory qualifier. **It is a governor-owned path** — including it would make this PR require owner ratification and block a `src/security/` change behind a governance review. Tracked on epic #5105 for the ratification batch instead.

### Verification

| mutation | result |
|---|---|
| drop `matchedRule` from the record | 2 fail |
| enforcer returns `unmeasured` before the denylist | 4 fail — incl. the previously-vacuous test |
| advisory branch stops recording | 4 fail — incl. the new sibling |
| remove the 500-char warning cap | 2 fail |

One process note: my first attempt at the third mutation reported "17 passed" — my regex had broken the file, so it proved nothing. Re-ran it targeted. A mutation row that changes the test *count* is not a kill, and it is worth checking the number rather than the colour.

`pnpm typecheck` clean, `eslint` clean, `markdownlint` clean on both docs, `vitest run src/security src/mcp`: **5791 passed, 0 failed**.

Issue #5125, opened 2026-08-28
six parallel implementations with split call sites: model selection, adapter acquisition, event bus, audit sink, task analysis, cost


williamzujkowski:
The six parallel implementations found in the 2026-08-27 audit, each with genuinely split non-test call sites. Filed as one issue so the corrected canonical table (#5121) has a specific reference, and so #5123 can seed its baseline from a single list rather than six scattered ones.

Verified against `main` v4.26.9. A shared interface with multiple implementors is polymorphism, not duplication — every entry below is two *separate abstractions* for one operation, with call sites divided between them.

## 1. Select a model for a task

- **A:** `cli-adapters/composite-router.ts:1088 createCompositeRouter` — 2 direct instantiations + 4 re-export surfaces
- **B:** `mcp/tools/delegate-to-model-helpers.ts:344 selectModel` + `scoreAllModels` — 1 internal site + **3 public re-exports** (`mcp/tools/index.ts:69`, `mcp/index.ts:322`, `exports/mcp.ts:129`)
- A has the 5-stage chain, circuit-breaker gating, outcome recording. B has a keyword-scored capability matrix, `model_hint` resolution, billing-mode suppression, and an availability-cache fallback A lacks.
- **CLAUDE.md declares A canonical.** B is undeclared and publicly exported. Aggravator: `cli/routing-audit-logic.ts:112` constructs `new TopsisRouter()` directly, which the same rule forbids.

## 2. Acquire an adapter

- **A:** `cli-adapters/factory.ts:115 createAllAdapters` — **7** non-test sites
- **B:** `adapters/unified-registry.ts:78 getGlobalRegistry` — **8** non-test sites
- A returns a raw `Map` with codex-transport plumbing. B adds process-wide caching, resilience wrapping, scoped resolution, lifecycle disposal.
- **The canonical table blesses BOTH** — `createAllAdapters()` under "CLI adapters", `UnifiedAdapterRegistry` under "Adapter registry" — while the Claude-specific section separately says access must go through the registry. The doc contradicts itself.

## 3. Emit a domain event

- **A:** `pipeline/event-bus.ts:53 getPipelineEventBus` — 7+ non-test sites
- **B:** `agents/collaboration/event-bus.ts:68 getGlobalEventBus` — 6+ non-test sites, and **re-exported as the core surface** by `core/event-bus.ts:28`
- A has pipeline stage events. B has topic-pattern subscriptions, correlation IDs, typed domain events, filters.
[...]

Issue #5123, opened 2026-08-28
concern registry + baseline ratchet: one canonical implementation per operation, adding a second one fails CI


williamzujkowski:
Child of epic #5121 (ratified: D, 5/6 approvers, 83.3%).

## What

A registry declaring **one canonical implementation per operation**, with a baseline-aware ratchet that fails CI when a new implementation of a registered concern appears.

## The unit is the operation, not the symbol

This is the whole point and the easiest thing to get wrong. CLAUDE.md's existing table lists symbols, which is why it can bless **both** `createAllAdapters()` and `UnifiedAdapterRegistry` — two entries for one question. Registry entries name an operation and its single entry point:

```yaml
- concern: acquire-an-adapter-for-a-model
  canonical: adapters/unified-registry.ts::getGlobalRegistry
  alternates:
    - path: cli-adapters/factory.ts::createAllAdapters
      callSites: 7
      trackedBy: '#TBD'
```

**Constraint 6 from the epic is binding:** if entries end up being module names rather than operations, this has reproduced the defect it was built to fix.

## Buy the detection; build only the wrapper

Constraint 1. Do **not** write a bespoke duplicate-detector. ESLint `no-restricted-imports`, `dependency-cruiser` forbidden-module rules, `knip`, and `jscpd` cover the detection. The novel part is the baseline diff — and that already exists in-tree three times (`check-tool-distinctness.ts`, `orphan-allowlist.json`, `schema-fanout-manifest.json`). Model the wrapper on those; it is a fourth instance of a proven pattern, not new CI architecture.

## Design against the merge-conflict failure

Constraint 2, and the contrarian's strongest objection. If the baseline stores exact call-site counts, two unrelated PRs touching an alternate both mutate it and the second hits a conflict — punishing developers who have nothing to do with the debt. A gate that does that gets disabled, and a disabled gate is worse than none.

Mitigations to evaluate, pick one and say why:
- `merge=union` in `.gitattributes` (already used for the governance ledgers)
- track **presence** of an alternate rather than a count
- track counts but only fail on *new paths*, refreshing counts automatically

## Seed entries

The eight known duplicates become the initial baseline, each linked to its tracking issue: middleware stacks (#4981), tool authorization (#5105), model selection, adapter acquisition, event bus, audit sink, task analysis, cost (#TBD — the pilot, which should be the first entry to reach zero alternates).

## Acceptance

[...]

Issue #5122, opened 2026-08-28
cost pilot: collapse eight cost paths onto one measured pipeline, with a PriceBasis that can say 'negotiated'


williamzujkowski:
Child of epic #5121 (ratified: D, 5/6 approvers, 83.3%). **The pilot.** Everything else in the epic is validated by whether this works.

## What

Collapse the eight cost paths onto one measured pipeline.

Current state, verified against `main` v4.26.9:

| # | Producer | Reaches registry? |
|---|---|---|
| 1 | `core/trace-pricing.ts:131 calculateCost` | yes — returns `undefined` when unpriced |
| 2 | `learning/usage-log.ts:108 computeCostDetail` | yes — returns `{costUsd:0, priced:false}` |
| 3 | `cli-adapters/budget-utils.ts:29 estimateCost` | yes — via `resolveCliCostPer1M`, **silently substitutes** a static rate |
| 4 | `testing/framework/test-metrics.ts:229 estimateCost` | yes, same substitution |
| 5 | `cli-adapters/budget-router-types.ts:101 DEFAULT_COST_MODELS` | **no** — hand-entered table whose own comment says it "deliberately differs" |
| 6 | `agents/orchestration/state-manager.ts:303` | **no** — hardcoded `costPerToken = 0.00001` |
| 7 | `workflows/aflow/evaluation-efficiency.ts:142` | **no** — sums step counts and timeout ms, named like money |
| 8 | `observability/decision-cost.ts rollupDecisionCost` | aggregation only |

Three lookups answer the same question three ways. **The unpriced-vs-zero distinction (#3855) exists in exactly one of them**, which is the property everything else should have inherited.

## Target

`computeCostDetail` is the honest one — it already carries `priced`, `resolvedId` and `matchedVia`. Consolidate onto it.

- [x] Paths 1–4 route through one function. Unpriced means one thing. **DONE** — path 1 #5174, paths 3+6 #5176, paths 2+4 #5178. Each unpriced policy is now named at its own call site (conservative fallback / fail-closed `undefined` / `priced:false`) rather than shared behind a flag.
- [ ] Paths 5–7 deleted, or renamed so they cannot be mistaken for money. **PARTIAL** — path 5 on the core via rate injection (#5181); path 6 deleted, now delegates to path 3 (#5176); path 7 remains and is a design decision, not a conversion (**#5180**). The aflow `estimateCost` naming point stands and is unaddressed.
[...]

Issue #5121, opened 2026-08-28
epic(arch): one canonical implementation per concern, and every measurement carries its basis — cost as the pilot


williamzujkowski:
Ratified architecture decision. Panel: 6/1 approve, option split **D:5 / A:1**, leading share **83.3%**, `thresholdMet: true`.

Prompted by the owner's observation that stubs and inconsistent implementations recur across the tree, and the ask for an approach that can "financially handle and grow to accommodate usage intelligently."

## What the evidence showed

Four independent audits against `main` v4.26.9. Everything below was verified with file:line, not inferred.

**Six NEW parallel implementations** with genuinely split call sites, beyond the two already tracked (#4981, #5105): model selection, adapter acquisition, event bus, audit sink, task analysis, cost. Two of them have *bridge modules* whose only job is shuttling between the duplicates.

**The canonical table is part of the problem.** CLAUDE.md lists **symbols, not concerns** — so it blesses both sides of adapter acquisition, names one event bus canonical while `core/event-bus.ts` re-exports the other as the core surface, and names a pricing source but no canonical cost function. `core/trace-pricing.ts:95` *admits in a comment* it is "a second, narrower implementation of the same lookup." Every violation is documented somewhere. **None fails a build.** That is the whole thesis.

**Cost is eight paths**, four reaching the model registry through three lookups with incompatible unpriced semantics, four never touching it at all — including a hardcoded `costPerToken = 0.00001` and one that sums step counts and timeouts and calls the result "cost."

**Eight instruments report structurally-constant values as measurements** (#5116, #5117, #5119).

**Three governance gates are dark today** (#5118).

## The decision

Build both mechanisms, with cost as the pilot that proves each:

- **A — concern registry + baseline ratchet.** One canonical implementation per *operation*, not per symbol. A committed baseline lists accepted alternates, each linked to its consolidation issue. CI fails on a new implementation of a registered concern.
- **B — provenance requirement.** Every reported measurement carries its basis (measured / estimated / unmeasured).

## Constraints from the dissent — these are binding, not decoration

The contrarian voted **reject** and the scope steward chose **A over D**. Both raised objections that survive the vote and shape how this gets built:

[...]

PR #5126, opened 2026-08-28
docs(governance): key the canonical-paths table on operations, not symbols

williamzujkowski:
**Owner-ratified in session** — *"I approve correcting the table."* Touches `AGENTS.md` and `CLAUDE.md`, both governor-owned, so the Governor-path ratification gate will show red until an approving review lands on this PR. Flagging that rather than merging past it silently.

Part of epic #5121 (ratified: option D, 5/6 approvers, 83.3%).

## The defect

The canonical-paths table listed **symbols**, not **operations**. That is how it blessed both sides of a duplication without anyone noticing:

- `createAllAdapters()` under "CLI adapters" **and** `UnifiedAdapterRegistry` under "Adapter registry" — two entries for one question, **7 call sites on one, 8 on the other** — while a separate line said adapter access must go through the registry. The document contradicted itself.
- The pipeline event bus named canonical, while `core/event-bus.ts` re-exports the *other* bus as the core surface, with two bridge modules shuttling between them.
- A pricing *source* named, but no canonical cost *function*. That is how eight cost paths accumulated.

Every duplication in #5125 was added in good faith by someone who could point at a blessed name.

## The change

Each row now answers **"what do I call to do X"**, with exactly one answer.

## `UNRESOLVED` is introduced as a real value

Three rows carry it: token→USD (#5122), domain events (#5125), audit records (#5125).

This is the same reasoning that produced `unmeasured` in #5100. Where two implementations exist and choosing between them is a **design decision rather than a cleanup**, the row says so and names the issue. A table that silently blesses both sides is worse than one that admits the fork — an author reading it cannot tell they are picking a side.

Each `UNRESOLVED` row carries interim guidance so the table stays actionable, and the surrounding prose says explicitly that the row is **not** to be resolved by editing the table. Take it to a panel, then record the decision.

## One row changes guidance rather than restating it

**Adapter acquisition** now names `getGlobalRegistry()` alone. The Claude-specific section already said this; the agnostic table contradicted it. This is the only row where the table's meaning changes rather than its shape — worth your eye.

## Mechanical notes

`AGENTS.md` is the source; `CLAUDE.md`'s block is generated. Both are in the diff because `pnpm governance:inject` regenerates the copy.

The governance check initially failed after my edit, and the cause is worth recording: `inject` applies table formatting that `check` does not, so a hand-formatted source table makes regeneration non-idempotent. Running `prettier --write AGENTS.md` **before** injecting fixes it. Anyone editing that table by hand will hit the same thing — the error message ("run `pnpm governance:inject`") does not hint at it, since running inject is exactly what does not help.

## Verification

- `pnpm governance:check` — exit **0** (verified as a real exit code, not a pipe's).
- `markdownlint` — 0 issues.
- Baselined first: `governance:check` passes on clean `main`, so the failure was mine and not pre-existing.

## Not in this PR

The ratchet that would *enforce* this table is #5123. A table is guidance; the enforcement is what changes the outcome. This PR makes the guidance correct so the ratchet has something true to enforce.

### DIFF
--- a/.changeset/canonical-table-by-operation.md
+++ b/.changeset/canonical-table-by-operation.md
@@ -1,0 +1,24 @@
+---
+'nexus-agents': patch
+---
+
+docs(governance): key the canonical-paths table on operations, not symbols
+
+The table listed important symbols, which let it bless both sides of a
+duplication without anyone noticing. It named `createAllAdapters()` under "CLI
+adapters" AND `UnifiedAdapterRegistry` under "Adapter registry" — two entries
+for one question, 7 call sites on one and 8 on the other — while a separate
+line said adapter access must go through the registry. It named the pipeline
+event bus canonical while `core/event-bus.ts` re-exports the other bus as the
+core surface. It named a pricing source but no canonical cost function, which
+is how eight cost paths accumulated.
+
+Each row now answers "what do I call to do X", with exactly one answer.
+
+`UNRESOLVED` is introduced as a real value. Where two implementations exist and
+choosing between them is a design decision rather than a cleanup, the row says
+so and names the tracking issue. A table that silently blesses both sides is
+worse than one that admits the fork: an author reading it cannot tell they are
+picking a side.
+
+Ratified by the owner. Part of epic #5121.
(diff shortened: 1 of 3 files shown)


williamzujkowski:

Issue #5134, opened 2026-08-29
research_synthesize declares four of the six fields it returns, so SDK-validating clients get -32602


williamzujkowski:
## What

`research_synthesize` fails with `-32602` on **every call that produces a result**. Its handler returns three properties its own `outputSchema` does not declare, and the MCP SDK enforces `additionalProperties: false`.

```
research_synthesize: MCP error -32602: Structured content does not match the
tool's output schema: data must NOT have additional properties,
data must NOT have additional properties, data must NOT have additional properties
```

Three violations, and they line up exactly.

| declared (`research-synthesize.ts:126-131`) | returned (`SynthesisResult`, `cli/research-helpers-synthesize.ts:150-157`) |
|---|---|
| `clusters` | `clusters` |
| `alignmentSummary` | `alignmentSummary` |
| `crossCuttingThemes` | `crossCuttingThemes` |
| `generatedAt` — *declared, never returned* | — |
| — | **`totalPapers`** |
| — | **`topicCount`** |
| — | **`featureGates`** |

## Why CI is green

`mcp-standalone-tools.test.ts` catches this locally and passes in CI, because CI's research registry is empty — the handler returns an error envelope rather than a `SynthesisResult`, so no structured content is validated. **The test only fails where the tool actually works.**

That is worth stating plainly: a green CI here means "the tool had nothing to return," not "the tool works." Anyone with a populated registry gets a hard `-32602`.

## The general defect, which is bigger than this tool

The schema carries this comment:

> Permissive shape — handler returns SynthesisResult (clusters, alignment, cross-cutting themes, etc.); inner shapes vary, **model the envelope only** (#2340 batch 3).

**"Model the envelope only" is not achievable under `additionalProperties: false`.** Declaring a subset of the returned keys is precisely what breaks — the SDK rejects anything undeclared. The pattern is self-defeating, and the comment reads as if it were a safe simplification.

Five other tools carry the same "permissive shape" construction: `memory-stats.ts`, `research-add-source.ts`, `research-analyze.ts`, `research-catalog-review.ts`, `research-discover.ts`. Each needs checking against its handler's real return type. Some pass today only because their return happens to match, which is luck rather than design.

## Fix

- [ ] Add `totalPapers`, `topicCount`, `featureGates` to `research_synthesize`'s `outputSchema`.
- [ ] Decide on `generatedAt` — declared but never returned. Either return it or drop it; a declared-and-absent optional field is a claim nothing supports.
[...]

Issue #5119, opened 2026-08-28
three instruments report structurally-constant values as measurements: security-gate triage, budget actualCostUsd, triangulated-review confidence


williamzujkowski:
Three findings from a stub-shaped-code audit, filed together because they share one shape: **a computed-looking number that is structurally constant, consumed by something that reports it as measured.** Each verified against `main` v4.26.9.

## 1. Security-gate triage returns a fixed verdict; `falsePositiveRate` is published as measured

`pipeline/security-gate.ts:128` — `const triageFn = config.triageFn ?? defaultTriageDelegate;`

`config.triageFn` has **zero producers**. Both production callers pass no config: `pipeline/agent-executor.ts:831` (`checkSecurityScan(target)`) and `mcp/tools/quality-gate-tool.ts:126` (`checkSecurityScan(projectDir)`). So `defaultTriageDelegate` (`:53-64`) always runs and returns a fixed `{confirmed: true, confidence: 0.5, suggestedSeverity: 'high'}` for every finding.

**Why it misreports:** `security-gate.ts:148/156` feeds `lifecycle.falsePositiveCount` — structurally always 0 — into `buildScanSummary`, which emits `"N filtered as false positives"` at `:284`, and `security/finding-lifecycle.ts:190` publishes `falsePositiveRate`. `getLastTriageVerdicts()` (`:46`) exports the constant verdicts as triage evidence.

A false-positive rate that is always 0 is not a good score; it is an absent measurement wearing one. Same class as the ClawGuard violation rate that was definitionally 100% (#5022).

## 2. Budget router debits an estimate and logs it as `actualCostUsd`

`cli-adapters/budget-router.ts:373` — `const actualCostUsd = result.value.costUsd ?? estimatedCostUsd;`

`CliResponse.costUsd` (`cli-adapters/types-core.ts:117`) is optional and **never assigned by any adapter** — neither `base-adapter.ts` nor `model-to-cli-adapter.ts:128-145` sets it; they set `usage` only. The `??` always takes the estimate.

**Why it misreports:** `:375` debits the session budget with it; `:381-386` logs `actualTokens`/`actualCostUsd`; and `cli/orchestrate-command.ts:196` surfaces `costUsd: (result.response as CliResponse).costUsd` — permanently `undefined` — in orchestrate output.

The name is the defect. `actualCostUsd` is a field that can only ever hold an estimate. Compounds the eight-path cost problem: this is a ninth consumer with its own fallback semantics.

[...]

Issue #5118, opened 2026-08-28
p1(governance): the gates read committed ledgers that no producer writes — 91 vote records exist in the runtime dir, 0 in the file the gates open


williamzujkowski:
## What

The governance gates read **committed** ledgers under `governance/`. The producers write to the **gitignored runtime** data dir under `.nexus-agents/governance/`. Nothing bridges them, so every gate that consumes a governance record aggregates over an empty set and reports clean.

Measured on `main` v4.26.9:

```
governance/vote-records.jsonl               0 bytes   since 2026-06-17
governance/pr-review-records.jsonl          0 bytes   since 2026-08-15
governance/authority-tier-transitions.jsonl 0 bytes   since 2026-08-26

.nexus-agents/governance/vote-records.jsonl 144,103 bytes / 91 records
                                            last written 2026-08-27 15:10
```

Those 91 records include this session's `higher_order` panels, the #5022 ClawGuard decision among them. **The records exist. The gates cannot see them.**

## Why — and it was deliberate, then deferred

`audit/vote-record-store.ts:59-65`:

> As of #3991 the runtime store no longer auto-writes here — this committed artifact is reached **only via the `NEXUS_VOTE_RECORDS_PATH` override** (the separate caller-commits/governor-gate path, **deferred**).

`resolveVoteRecordsPath()` (`:351`) honours the env override when set, otherwise resolves `nexusDataPath('governance', 'vote-records.jsonl')` — the per-repo runtime dir.

**Nothing sets the override.** `grep -rn NEXUS_VOTE_RECORDS_PATH .github/ scripts/ package.json` → no hits.

So #3991 split producer from consumer on purpose and deferred the bridge. The deferral lives in a JSDoc comment. Per the repo's own rule — *memory notes, PR-description follow-up bullets, and code TODOs are NOT tracking* — that is exactly the case this issue exists to correct.

## What is dark as a result

- **`check-governor-review.ts:56`** reads `governance/pr-review-records.jsonl`. Empty → the gate takes its warn-first "no diff-bound record" branch and exits 0 on every governor-path PR. This is the gate CLAUDE.md relies on so the governor cannot weaken its own governor.
[...]

Issue #5117, opened 2026-08-28
p1(governance): proof_of_learning and QuorumValidator report weighted tallies where every weight is structurally 1.0


williamzujkowski:
## What

Two consensus surfaces branch into a "weighted" algorithm, report weighted tallies, and describe themselves to users as performance-weighted — while every weight is structurally `1.0`.

### 1. `proof_of_learning` strategy

`consensus/engine.ts:381` sets each voter's weight from `calculateVoteWeight(performance)`. The performance map (`engine.ts:130`) is populated **only** by `updateAgentPerformance` (`engine.ts:282`), which has **zero non-test callers** — verified: a repo-wide grep for the symbol outside tests returns exactly one hit, its own definition.

So `performance` is always `undefined`, and `strategies.ts:298-300` returns the new-agent default:

```ts
if (performance === undefined || performance.totalVotes === 0) {
  return 1.0; // Default weight for new agents
}
```

At the default threshold this is arithmetically identical to `simple_majority`.

**The misreport:** `strategies.ts:286-289` returns `reason: "Approved with X% weighted approval"` and a populated `weightedCounts` field. The MCP tool schema advertises it to callers as *"proof_of_learning: Weighted by agent performance (Issue #103)"* (`mcp/tools/consensus-vote-types.ts:38`, user-visible enum description at `:217`).

### 2. `QuorumValidator`

`consensus/quorum-validator.ts:244` — `let weight = agentWeights?.get(agentId) ?? 1.0;`

`QuorumValidationInput.agentWeights` (`:59`) has **no non-test producer**; every path through `validateQuorum` (`:158-164`) passes `undefined` through. `calculateWeightedQuorum` (`:268-270`) still takes the weighted branch and emits reasoning text naming weighting, for what is a plain headcount.

## Why this is p1 rather than cosmetic

This is a **governance instrument on the governor path**. Per CLAUDE.md, an instrument that records a decision must be able to represent what it measures, and a fidelity defect there is a p1 correctness bug rather than a documentation nit.

A vote record saying *"approved with 71% weighted approval"* invites a reader to believe voter track record influenced the outcome. It did not, and cannot, because nothing has ever written a track record. This is the same class as #4986 (an `absolute_quorum` void persisting as `approved`) — the record describes a mechanism that did not run.

It also interacts with #4943: the persisted voter entry already drops which adapter each voter ran on. A reader cannot see that three of seven voters shared a model **and** is told the tally was performance-weighted.

[...]

PR #5136, opened 2026-08-29
docs(rules): trace to the producer, and check what a pass measured

williamzujkowski:
**Owner ratification required** — `.rules/` is governor-owned (`CODEOWNERS:23`, above the sentinel). I will not self-merge.

Ratified **5/6, 83.3%, supermajority met** by a `higher_order` panel.

## What this adds — and what it deliberately does not

I measured the overlap before proposing. `.rules/debugging.md` already covers Stop-the-Line, the Triage Sequence, non-reproducible-bug tactics, and an Anti-Rationalization table that includes *"I know what the bug is"*, *"The failing test must be wrong"*, *"It works on my machine"* and *"This is flaky, ignore it"*. `development-disciplines.md` covers "name the empty case".

**Most of what this session used was already written down.** So this is **two table rows and one triage step** — about 12 lines — not a new section. The panel made that a condition: *"the value is salience, not exposition."*

### 1. Trace to the producer, not the guard

The word "producer" appeared in **no rule file**. Yet the dominant defect class found this session is always one shape — the check reads correctly and nothing upstream can make it fire:

- ClawGuard deciding on an `allowedTools` no producer ever populates (#5022)
- `proof_of_learning` reporting weighted approval where `updateAgentPerformance` has zero non-test callers (#5117)
- `falsePositiveRate` published as measured from a constant triage verdict (#5119)
- Three governance gates reading ledgers no producer writes (#5118)

"Name the empty case" is adjacent, but governs *writing* code. This governs *investigating* it.

### 2. A pass is evidence only once you know what it measured

Distinct from "It works on my machine", which is about environment diffs. And directionally **opposite** to the flaky row: that one says don't ignore a failure; this says don't trust a pass.

- A CI gate run locally that passed **because it ran on the wrong branch** — `HEAD` didn't match the PR head
- A poll loop reporting SETTLED from the **previous** run's results
- A seam test passing because the parameter was named `template`, not `workflow`, so validation rejected the call before it reached the code under test
- **#5134** — `research_synthesize` fails `-32602` on every real call, and CI is green *only* because CI's registry is empty, so the tool returns nothing to validate

That last one is the argument. **The test only fails where the tool actually works**, and I nearly dismissed it as pre-existing noise.

## The dissent, recorded rather than dismissed

The contrarian voted reject and called this the *"documentation placebo fallacy"*:

> The ~15 failures cited are infrastructure and tooling failures, not documentation gaps. If a test passes because the tool returns an error envelope and no structured content is validated, the test runner is fundamentally broken. We should be writing strict linters rather than asking agents to second-guess green CI runs. Incremental verifiability: zero.

**That is substantially right, and several approvers conceded it** — rules are not measurable, and the enforcement fix is epic #5121's ratchet, not prose. I have not claimed otherwise: this is the investigation posture, the ratchet is the enforcement, and they are complements.

The specific engineering fix the contrarian points at is already tracked — #5134's acceptance requires the round-trip test to run against a **populated** registry, so it stops certifying tools that only work when they return nothing.

**What decided it against option D** (put these in `skills/bug-fix` instead), raised independently by three voters: **skills are pull-based.** An agent opens the bug-fix playbook once it knows it's debugging. The failure mode here is *not knowing there is a bug*, because the pass looked like evidence. That needs the auto-loaded surface.

## A correction I made mid-flight

I told the owner there was a gap — that `CLAUDE.md` lists `.rules/` as governor-owned while `CODEOWNERS` omits it. **That was wrong.** `.rules/` is at `CODEOWNERS:23`, above the sentinel; the gate does cover it. I'd asserted it from recalling the governor block instead of reading the file, and would have filed a false issue. Caught by measuring before filing — which is item 1 of this very PR.

## Verification

`markdownlint` clean, `governance:check` **exit 0** (verified as a real exit code, not a pipe's).

### DIFF
--- a/.changeset/debugging-producer-and-pass-evidence.md
+++ b/.changeset/debugging-producer-and-pass-evidence.md
@@ -1,0 +1,26 @@
+---
+'nexus-agents': patch
+---
+
+docs(rules): trace to the producer, and treat a pass as evidence only once you know what it measured
+
+Two investigation techniques added to `.rules/debugging.md`, ratified 5/6 by a
+`higher_order` panel. Overlap with the existing rules was measured first: most of
+what this session used is already covered by the Triage Sequence and the
+Anti-Rationalization table, so the addition is deliberately two table rows and
+one triage step, not a section.
+
+**Trace to the producer, not the guard.** The word "producer" appeared in no rule
+file, yet the dominant defect class found this session is one shape: the check
+reads correctly and nothing upstream can make it fire. ClawGuard deciding on an
+`allowedTools` no producer populates; `proof_of_learning` reporting weighted
+approval where `updateAgentPerformance` has zero non-test callers;
+`falsePositiveRate` published as measured from a constant verdict; three
+governance gates reading ledgers no producer writes.
+
+**A pass is evidence only once you know what it measured.** Distinct from the
+existing "It works on my machine" row, which is about environment diffs. This is
+the opposite direction from "This is flaky, ignore it": that row says do not
+ignore a failure, this says do not trust a pass. It is the direction that hid
+#5134 — `research_synthesize` erroring on every real call while CI stayed green,
+because CI's registry is empty and the tool returned nothing to validate.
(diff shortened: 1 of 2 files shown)


williamzujkowski:

